PR #25, opened 2022-08-12
Sphinx conf.py fixes

mavaylon1:
There were some issues with the conf.py that would make the build (on read the docs) fail.

I've made the necessary changes and added the pdf version of the logo for latex.

I can confirm that build does work with these changes.


oruebel (commented):
Can you please remove the ``.DS_Store`` files and add it to the list in ``.gitignore``. Otherwise the changes look good.

PR #27, opened 2022-08-12
Create run_flake8.yml

mavaylon1:
Created flake8 workflow.

### DIFF
--- a/.github/workflows/run_flake8.yml
+++ b/.github/workflows/run_flake8.yml
@@ -1,0 +1,29 @@
+name: Run style check
+on: pull_request
+
+jobs:
+  run-flake8:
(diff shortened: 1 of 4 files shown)


oruebel:
For falke8 the settings are in ``setup.cfg``. For the gallery, I would add `` docs/gallery/*:E402,T001`` to the ``per-file-ignores=`` since in the gallery we usually don't want to place all module level imports on top of the file (i.e., we want to ignore E402). You can also have a look at https://github.com/hdmf-dev/hdmf/blob/dev/setup.cfg for reference

rly:
Can you also fix the flake8 errors, in this or a separate PR?

rly (approved):

Issue #18, opened 2022-08-11
Setup tox for running unit tests


oruebel:
See https://github.com/hdmf-dev/hdmf/blob/dev/tox.ini for an example

oruebel:
This issue has been fixed in #32

Issue #11, opened 2022-08-11
Setup branch protections


oruebel:
Setup GitHub branch protections for the dev branch (similar to the setup in HDMF) to: i) require PRs and prevent direct commits to the dev branch, ii) require that PRs pass the main CI checks (see #10)

oruebel:
@mavaylon1 thanks for setting up the branch protections. In addition to the current setup we should also specify the status checks that are required to allow merge (e.g., flake8, minimum coverage, pass main tests etc.)

Here how this looks in HDMF: 

![Screen Shot 2022-08-19 at 4 07 33 PM](https://user-images.githubusercontent.com/10999845/185718011-1ba41943-9e86-4ae1-8673-cf424ad1220e.png)

Issue #10, opened 2022-08-11
Setup GitHub Actions/Pipelines for automatic CI


oruebel:
Similar to the main HDMF we should run the following all pull requests
- flake8 (or black)
- unit tests
- gallery tests for sphinx gallery as part of the docs
- docs build
- linkcheck on docs
- May Require #18 (but not sure)

For this we can most likely build off the workflows defined in HDMF https://github.com/hdmf-dev/hdmf/tree/dev/.github/workflows

bendichter:
+1 for black. See how we set it up here:

https://github.com/NeurodataWithoutBorders/nwbinspector/blob/803756995f84d04bf543c85fb1684961e6290643/.pre-commit-config.yaml

https://github.com/NeurodataWithoutBorders/nwbinspector/blob/60e081c68710308c05a48f6b2211480b38e98589/pyproject.toml

https://github.com/NeurodataWithoutBorders/nwbinspector/blob/bac7056cc5f2c61a1444b5a6c8a9e990a5c26b4c/.github/workflows/black.yml

oruebel:
> +1 for black

👍 I have not used black before, but I won't stand in the way of progress 😉

oruebel:
Fixed in #32

PR #32, opened 2022-08-16
Create workflows

mavaylon1:


mavaylon1:
Fix #18, #10 , #11

docs/source/overview.rst
```diff
@@ -42,4 +42,4 @@ TODO
...
 - Handling of  external links on export is not yet fully implemented and is missing a few corner cases
-- Here we update the PyNWB test harness to add ZarrIO to the rountrip tests, which in turn runs all HDF5 roundtrip tests also for Zarr. This requires changing the test harness in PyNWB, instead it would be useful to be able to "inject" new I/O backends in the test harness so that we can specify those tests here, rather than implementing this in PyNWB and making PyNWB dependent on hdmf-zarr (see https://github.com/NeurodataWithoutBorders/pynwb/pull/1018/files#diff-ebaf216d6ea84b72ee9ac17ab17b6e5d5ed8a5f3bbd09ee11d9ee6969c2fc294 )
\ No newline at end of file
+- Here we update the PyNWB test harness to add ZarrIO to the rountrip tests, which in turn runs all HDF5 roundtrip tests also for Zarr. This requires changing the test harness in PyNWB, instead it would be useful to be able to "inject" new I/O backends in the test harness so that we can specify those tests here, rather than implementing this in PyNWB and making PyNWB dependent on hdmf-zarr (see https://github.com/NeurodataWithoutBorders/pynwb/pull/1018/files)
```
oruebel: The main reason I had used the other link is because the main changes this TODO item is referring to are the changes in ``tests/integration/ui_write/base.py`` as part of that PR (as well as the changes in ``tests/integration/hdf5/test_modular_storage.py``). The other changes in that PR (i.e., the additions for ``docs/notebooks/zarr_file_conversion_test.ipynb`` and ``src/pynwb/__init__.py`` are already part of the hdmf-zarr library. However, I can see that using the more specific link is confusing and the intent is not clear. Maybe we should just update this note to clarify what the relevant files from this PR are..

docs/source/overview.rst
```diff
@@ -41,4 +41,5 @@ TODO
...
\ No newline at end of file
+=======
+- Here we update the PyNWB test harness to add ZarrIO to the rountrip tests, which in turn runs all HDF5 roundtrip tests also for Zarr. This requires changing the test harness in PyNWB, instead it would be useful to be able to "inject" new I/O backends in the test harness so that we can specify those tests here, rather than implementing this in PyNWB and making PyNWB dependent on hdmf-zarr. See the files ``tests/integration/ui_write/base.py`` and ``tests/integration/hdf5/test_modular_storage.py`` as part of `PyNWB #1018 <https://github.com/NeurodataWithoutBorders/pynwb/pull/1018/files>`_ for details (the other parts of the this PR have already been ported to *hdmf-zarr*).
```
oruebel: The ``=======`` in Line 44 looks like it may be a leftover from a merge conflict and should be removed

test.py
```diff
@@ -0,0 +1,148 @@
...
+import re
+import argparse
+import logging
+import os.path
+import os
+import sys
+import traceback
+import unittest
+
+flags = {'hdmf': 1, 'example': 4}
```
oruebel: Instead of ``'hdmf'`` I think we should rename to the flag to ``'unit'``

oruebel:
This looks good to me. @rly could you review this PR since you are more familiar with the CI.

.github/workflows/black.yml
```diff
@@ -0,0 +1,13 @@
+name: black
+
+on: pull_request
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
```
rly, line 9: These actions have newer versions. Can you update these?

rly:
I recommend keeping `requirements-min.txt` from HDMF and testing with it. They should mirror the minimum requirements set in `setup.py`. We should test those in the CI (please revert some of the changes where you removed py3.7-minimum)

mavaylon1:
@oruebel So the the gallery tests for 3.7 minimum are using an older version of dandi. I've tried to set a dandi minimum version but that breaks other parts. 
ex)
I added dandi>= 0.46.1 to the requirements-doc.txt which led to 
1. Run code coverage / macos-latest failing when it was passng prior.
2. The same gallery 3.7 minimum tests failing but for a new reason. ----> "FileNotFoundError: [Errno 2] No such file or directory: 'requirements.txt' ". 

For more details: checkout commit df51116

Any thoughts?

This is for the doc "plot_convert_nwb.py" line 19 showing an error in the init that seems to be fixed in newer dandi versions.

oruebel (approved):
Looks good to me

PR #51, opened 2022-12-16
Fix Github worklfow/tox

mavaylon1:


rly:
I think I figured out the failing conda linux 3.7 test. It has to do with pip and virtualenv not being updated. See this workaround: https://github.com/tox-dev/tox/issues/1768#issuecomment-754478737

While you are updating this workflow @mavaylon1, can you also apply the updates that I made here to the HDMF package: https://github.com/hdmf-dev/hdmf/commit/ccc5c252db697a65439086ff925f729e5eae118d

oruebel (dismissed):

.github/workflows/run_coverage.yml
```diff
@@ -54,9 +54,10 @@ jobs:
         run: |
           python -m pip install -e .  # must install in editable mode for coverage to find sources
           python -m pip list
+          python -m pip install --upgrade numpy
```
rly: Is this necessary? It is not part of the other workflows. If a higher version of numpy is required, then we should specify that elsewhere.

requirements-min.txt
```diff
@@ -3,3 +3,4 @@ zarr==2.11.0
 numcodecs==0.9.1
 pynwb==2.0.0
 setuptools
+numpy==1.19.3
```
rly: Could you please add a comment describing why this is the minimum version?

Also if this is the minimum version, it should also be specified in the `reqs` variable in `setup.py`

.github/workflows/run_coverage.yml
mavaylon1: I did specify a higher version in the requirements file but it didn't work. But I can try again given the updated workflows. Updating the numpy version was the fix for the codecov windows test.

rly:
I left a couple comments to address but otherwise it looks good to me.

.github/workflows/run_coverage.yml
rly: Yes, please try it again. If hdmf-zarr needs a higher numpy requirement than what is specified by the other packages in order to run, then it should be specified in `setup.py` and not only in this workflow.

rly (dismissed):

rly:
Looks good to me. Thanks!

rly:
One more minor thing: can you update the `black` workflow to use the latest versions of the used actions?

rly (approved):

Issue #52, opened 2022-12-22
Update tox.ini to use test_gallery.py and fix gallery-python-3.7 tests


mavaylon1:
1. Update tox.ini to use test_gallery.py to be in line with HDMF
2. Currently both linux-gallery-python3.7-minimum and windows-gallery-python3.7-minimum will pass locally when running "python test.py --example", but not during the github checks. I've also tested a version of test_gallery.py by running in a branch "python test_gallery.py"; however, this returns an error regarding missing files. (Refer to attached images)
![Screen Shot 2022-12-21 at 11 01 36 AM](https://user-images.githubusercontent.com/22578631/209206869-e13107b7-5138-486b-8cc7-e7c3fe7af017.png)

mavaylon1:
@rly I'd love some help with this. I've tested all my ideas of checking if the the files are indeed found in the directory by printing out the list of files used in test_gallery.py, and fiddling with absolute paths.

rly:
I'll take a look

rly:
Closed with #52

Issue #45, opened 2022-12-13
Arrays possibly being transposed when converting NWB files from HDF5 to ZARR


oruebel:
The tutorial for converting NWB data from HDF5 to Zarr, currently shows the following warnings (see also See https://hdmf-zarr.readthedocs.io/en/latest/tutorials/plot_convert_nwb.html#read-the-zarr-file-back-in)

<img width="952" alt="Screen Shot 2022-12-13 at 12 31 26 AM" src="https://user-images.githubusercontent.com/10999845/207265575-17469f77-908f-46ba-adcd-68275d947f6e.png">

In particular the warning ```warn("Length of data does not match length of timestamps. Your data may be transposed. Time should be on "``` is something that should be looked at, as it appears that (some) arrays may for some reason be transposed in the conversion.

oruebel:
Using the latest version of PyNWB and HDMF locally, the warning we see currently with ```warn("Length of data does not match length of timestamps. Your data may be transposed. Time should be on "``` does not occur, but instead the following warnings are now being raised: 


```
/Users/oruebel/Devel/nwb/pynwb/src/pynwb/core.py:47: UserWarning: OpticalSeries 'StimulusPresentation_encoding': The number of frame indices in 'starting_frame' should have the same length as 'external_file'.
  warn(error_msg)
/Users/oruebel/Devel/nwb/pynwb/src/pynwb/core.py:47: UserWarning: OpticalSeries 'StimulusPresentation_encoding': Either external_file or data must be specified (not None), but not both.
  warn(error_msg)
```

Reading the data directly from HDF5 yields the same warnings, i.e., these warnings are actually due to errors in the file that is being used in the tutorial rather than indicating and error in the library.

oruebel:
I also tested in the regular tutorial, and it looks like arrays are  indeed not being transposed. It appears that the warnings currently may be due to versions of PyNWB/HDMF used, that have already been addressed on the dev branches of those libraries. The new warnings for ``OpticalSeries`` are correct and are due to errors in the file that is being used. 

Closing this issue as it does not appear to be an issue in HDMF_ZARR.

PR #53, opened 2022-12-22
Add test_gallery.py and debug python_3.7 gallery tests

mavaylon1:


mavaylon1:
#52

mavaylon1:
1. Even though we are switching the gallery tests to test_gallery.py, the current use of "python test.py --example" leads to different results. On my local conda env with python 3.7, "python test.py --example" passes; however, here it returns multiple "File not Found" errors.
2. Running "python test_gallery.py" returns the following error for "plot_convert_nwb_hdf5.py" locally (refer to image). This is a user warning, should I add an exception? @oruebel 
Screen Shot 2022-12-22 at 11 49 36 AM

oruebel:
> 2\. Running "python test_gallery.py" returns the following error for "plot_convert_nwb_hdf5.py" locally (refer to image). This is a user warning, should I add an exception?

The user warning that ``The number of frame indices in `starting_frame`` should have the same length as ``external_file`` I think may be the same as this issue https://github.com/hdmf-dev/hdmf-zarr/issues/45  However, it could also be a bug in the tutorial. It would be nice to raise errors in the gallery test when unexpected warnings occur, however, I'm not sure right now how to configure this, since there may be some warnings that are expect to occur.

oruebel:
> 2\. Running "python test_gallery.py" returns the following error for "plot_convert_nwb_hdf5.py" locally (refer to image). This is a user warning, should I add an exception?

The strange thing is though, that this particular warning does not seem to happen on ReadTheDocs https://hdmf-zarr.readthedocs.io/en/dev/tutorials/plot_convert_nwb_hdf5.html Which version of PyNWB are you using? I'm just wondering whether this a new check on the dev branch of PyNWB so that you are seeing it on your system but it does not yet appear when using the release version of PyNWB.

rly:
This line at the end of `plot_nwb_zarrio.py`
https://github.com/hdmf-dev/hdmf-zarr/blob/6841196aac0b8d66a112bedf4fc0f26b6de48272/docs/gallery/plot_nwb_zarrio.py#L155

changes the current directory of the Python execution which messes up finding the other gallery files. 

I would suggest changing the current directory back at the end of that script.

This does not happen in `test_gallery.py` I think because the `_import_from_file` function is different and more robust to changes in file location. 

@oruebel The warning was added in PyNWB 2.1.1: https://github.com/NeurodataWithoutBorders/pynwb/pull/1516
The `requirements.txt` file in this repo pins PyNWB to version 2.0.1 and this is used by readthedocs.

rly:
I think it is fine to move to using `test_gallery.py` for gallery tests and `pytest` for all other tests

oruebel:
> changes the current directory of the Python execution which messes up finding the other gallery files

Good catch! To be honest, this part of the script

https://github.com/hdmf-dev/hdmf-zarr/blob/6841196aac0b8d66a112bedf4fc0f26b6de48272/docs/gallery/plot_nwb_zarrio.py#L144-L158

should actually be changed to a unit instead. The reason for changing the current directory here was because I was debugging the issue with links and changing the current directory was a way to trigger the issue without copying files. For now, I would suggest the following: 1) Remove those lines from the gallery, 2) Open an issue to add a test for this case, basically trimming down the notebook and making it into a unit test, and 3) actually implement the unit test.

oruebel:
I have now removed the problematic lines from the ``plot_nwb_zarrio.py`` tutorial and create a new issue https://github.com/hdmf-dev/hdmf-zarr/issues/58 to add a test to cover this case in the regular test suite. It looks like with this change, the 3.7 pipelines are not passing.

mavaylon1:
If I downgrade to pynwb 2.1.0 the gallery tests pass. This seems to be a new warning/error coming from pynwb 2.2.0. For now I can downgrade pynwb to be greater than 2.0.0 but less than or equal to 2.1.0. That way we can merge test_gallery and look create a ticket for debugging the tutorials to whatever pynwb 2.2.0 needs. @oruebel

oruebel:
> . This seems to be a new warning/error coming from pynwb 2.2.0.

Just to clarify, are you referring to the following warning:

```
UserWarning: OpticalSeries 'StimulusPresentation_encoding': The number of frame indices in 'starting_frame' should have the same length as 'external_file'.`
```

or are there other warnings/errors? This particular warning is due to an error in the NWB file from DANDIset ``000207`` we are using in the tutorial:

https://github.com/hdmf-dev/hdmf-zarr/blob/6841196aac0b8d66a112bedf4fc0f26b6de48272/docs/gallery/plot_convert_nwb_hdf5.py#L17-L18

The file is using the ``starting_frame`` field wrong. The correct usage for this field was clarified in the schema and API but I assume this file was created before than and used the field wrong. This is now raises a warning in PyNWB. To address this issue we should probably use a different dataset in the tutorial. Rather than downloading a dataset from DANDI it may be best to use the PyNWB test/mock function to create a dummy NWB file and use that instead in the test.

mavaylon1:
@oruebel 
My bad I thought this issue was in the chain of comments above. I was referring to the image attached here for plot_convert_nwb.py
Screen Shot 2023-01-02 at 5 00 34 PM

mavaylon1:
So as of now I've set pynwb to 2.1.0 and downgraded numpy to make compatible with pynwb setup requirements. The remaining error is sphinx, but this should be good to go for a release with the idea of getting something updated/working out there, with the next release geared towards "using the latest and greatest from pynwb". I'll keep working on sphinx but let me know what you think. @oruebel

mavaylon1:
Fixed sphinx to an older version currently being used by hdmf.

oruebel:
@mavaylon1 https://github.com/hdmf-dev/hdmf-zarr/pull/56 fixes the issue in conversion tutorial and another bug in the ZarrIO tutorial.

mavaylon1:
@oruebel so we need the gallery test for 3.10 to pass. Let's merge this with pynwb==2.1.0, and then change it to be more general in #56.

mavaylon1:
I'll take a look at the windows gallery since it is throwing an error not related to the tutorial you updated. However, the sphinx test for external links is throwing an error due to the updated conversion tutorial. @oruebel

oruebel:
@mavaylon1 does the linkcheck run a different setup? It is strange that the tutorial is failing there but in none of the other checks. I'll take a look later today.

docs/gallery/plot_convert_nwb_hdf5.py
```diff
@@ -60,6 +60,8 @@
 with NWBHDF5IO(filename, 'r', load_namespaces=False) as read_io:  # Create HDF5 IO object for read
     with NWBZarrIO(zarr_filename, mode='w') as export_io:         # Create Zarr IO object for write
         export_io.export(src_io=read_io, write_args=dict(link_data=False))   # Export from HDF5 to Zarr
+        export_io.close()
+    read_io.close()
```
oruebel: We should not need to call ``close`` on the IO objects explicitly.  ``close`` is being called by the ``__exit__`` function here:

https://github.com/hdmf-dev/hdmf/blob/23a8a2638f060366ec1e7789f1344c9732e42e43/src/hdmf/backends/io.py#L133-L134

when we leave the context

oruebel:
It looks like some of the failures are due to DANDI download. Since the file is small, we could just add it to the repo and not download from DANDI here.

docs/gallery/plot_convert_nwb_hdf5.py
oruebel: I removed extraneous io.close calls in #61

oruebel (approved):

PR #57, opened 2022-12-24
Store object_id with links if available

### LINKED ISSUE #54: Save object id's as part of links and references
- To enhance portability of links and references it would be nice to store the object_id of the Zarr file in addition to the relative path when a link/reference points to an external file. This will be useful both for error checking but can also help resolve links in case that path's are not valid

oruebel:
Fix #54 

This PR updates the storage of links/references to add the following information:

- ``object_id``: Object id of the reference object. May be None in case the referenced object does not have and assigned object_id (e.g., in the case we reference a dataset with a fixed name but without and assigned data_type (or neurodata_type in the case of NWB).
- ``source_object_id``: Object id of the source Zarr file indicated by the ``source`` key. The ``source`` should always have an ``object_id`` (at least if the ``source`` file is  a valid HDMF formatted file).

TODO:
- [x] Updated the ``ZarrReference`` class to add a ``source_object_id`` and ``object_id`` keys
- [x] Updated ``ZARRIO.__get_ref`` to populate the ``source_object_id`` and ``object_id`` keys
- [x] Updated the storage documentation to document the  ``source_object_id`` and ``object_id`` keys and update examples
- [x] Update CHANGELOG
- [x] Update tests


tests/unit/test_io_zarr.py
```diff
@@ -877,9 +891,16 @@ def test_soft_link_group(self):
...
             write_io.write(foofile)
+        with open(self.paths[0]+"/.zattrs", 'r') as f:
+            print(f.readlines())
+
         with ZarrIO(self.paths[0], manager=get_foo_buildmanager(), mode='r') as read_io:
             with ZarrIO(self.paths[1], mode='w') as export_io:
-                export_io.export(src_io=read_io, write_args=dict(link_data=False))
+                with self.assertWarnsWith(UserWarning, "Could not determine source_object_id for "
+                                                       "builder with path: /buckets/bucket1/foo_holder/foo1"):
+                    export_io.export(
```
oruebel: Check if this warning is really expected to be raised (e.g., which is the case if the file is not using the ROOT_NAME as the name of  the root builder or has no object_id assigned). Otherwise we should not see this warning, and we may need to do something different during export. Just something to double check.

oruebel:
@mavaylon1 we should check whether this PR now also works with the fixes in #120 It would be nice we could include this in the release as well if it works. Otherwise, its fine to move this PR to the next release, but would be nice to push this over the finish line.

tests/unit/test_io_zarr.py
oruebel: This should be resolved by #120 which names FooFile correctly as `root`

oruebel:
Aside from adding/updating unit tests to check that the values for the `object_id` and `source_object_id` fields are correct, this PR should be ready.

oruebel:
> adding/updating unit tests to check that the values for the `object_id` and `source_object_id` fields are correct

Done

mavaylon1:
@oruebel I wanted to as about the case when the object_id is None. You said that would be the case when the data_type/neuro_datatype is not assigned for a dataset. I was thinking of an example of what that would be. Say we have TimeSeries, which is a dataset. This has a type, but it also contains a dataset "data" that does not. This would be an example where we have "object_id" as none if that was the target correct?

oruebel:
> . This has a type, but it also contains a dataset "data" that does not. This would be an example where we have "object_id" as none if that was the target correct?

Correct. For `TimeSeries.data` the `object_id` would be `None` because it is just a dataset within a type. However, the `source_object_id` should always be present since the file is always represented by a Container. The `object_id` and `source_object_id` are not really being used right now, but will be useful to validate links and possibly in the future to be able to retrieve external links dynamically.

mavaylon1 (approved):

PR #62, opened 2023-01-05
[Feature] Add support for using select user-defined zarr stores

oruebel:
# Motivation

Currently ZarrIO only supports Zarr's the default DirectoryStore. It would be nice to add support for additional data stores.

- [X] Allow custom ``DirectoryStore``, ``NestedDirectoryStore`` and ``TempStore`` for read and write in ``ZarrIO``
- [X] Update ``test_zarr_io.py`` to support specification of backend stores and add unit tests based on ``test_zarr_io.py`` for the new stores. 
- [X] Update handling of references on read to support file-based stores (e.g., ZipStore or database stores. This will require updating the read logic to open target object not via the full path, but by opening the Zarr file itself and then retrieving the linked object. 
- [x] Update NWBZarrIO to support the new options of ZarrIO
- [x] Add export tests based on ``test_io_convert.py`` for the new stores
- [x] Add developer documentation for how to integrate new storage backends with ``ZarrIO``
- [x] Update ZarrIO tutorial (and maybe the NWB tutorial) to show how to write using custom Zarr stores


oruebel:
I will split this into two PRs. Allowing TempStore, DirectoryStore, and NestedDirectoryStore is much simpler as those all derive from DirectoryStore. Adding the more file- and database-oriented stores, e.g., zip or SQL, will be more work with regard to support for links and references (especially during export and when linking across Zarr files with different backend stores). I created issue #65 to explore adding more stores.

oruebel:
@rly @mavaylon1 this PR is ready for review

oruebel:
Note, the external link check is failing here, because it includes a link to a file that is being added to the main repo by this PR. I.e., once this PR is merged that link should be correct.

tests/unit/test_io_convert.py
```diff
@@ -34,28 +43,58 @@
...
 
 class MixinTestCaseConvert(metaclass=ABCMeta):
     """
     Mixin class used to define the basic structure for a conversion test.
+
+    To implement a test case using this mixin we need to implement the abstract methods
+    ``setUpContainer`` and ``roundtripExportContainer``. The behavior of the mixin can
+    then be further customized via the following variables:
+
+    :ivar IGNORE_NAME: Bool parameter passed to assertContainerEqual (False)
```
rly: note that these variables are actually class variables, not instance variables, and I think `:ivar name:` will not work here. Also, as a class variable, if one test changes the mixin class, I think it will change for all tests that use the class downstream. It seems like the mixin class is never instantiated directly though, and these variables are modified only in the subclass definition, so that should be OK [edited]

tests/unit/test_io_convert.py
```diff
@@ -65,62 +104,90 @@ def tearDown(self):
...
+                # make sure we get a completely new object
+                self.assertNotEqual(id(container), id(exported_container), message)
+                # the name of the root container of a file is always 'root' (see h5tools.py ROOT_NAME)
+                # thus, ignore the name of the container when comparing original container vs read container
+                self.assertContainerEqual(container,
+                                          exported_container,
+                                          ignore_name=self.IGNORE_NAME,
+                                          ignore_hdmf_attrs=self.IGNORE_HDMF_ATTRS,
+                                          ignore_string_to_byte=self.IGNORE_STRING_TO_BYTE)
+                # TODO: Add message to the assertContainerEqual call once the feature is released in HDMF
```
rly: this is about to be released in HDMF 3.5.0

tests/unit/base_tests_zarrio.py
```diff
@@ -502,60 +567,79 @@ def test_write_attributes_write_3Darray_of_floats(self):
...
+        tempIO = ZarrIO(self.store, mode='w')
+        tempIO.open()
+        testgroup = tempIO.file  # For testing we just use our file and create some attributes
         attr = {'attr1': dataset_1}
-        self.io.write_attributes(testgroup, attr)
+        tempIO = ZarrIO(self.store, mode='w')
+        tempIO.write_attributes(testgroup, attr)
         expected_value = {'attr1': {'zarr_dtype': 'object', 'value': {'source': ".", 'path': '/dataset_1'}}}
         self.assertDictEqual(testgroup.attrs.asdict(), expected_value)
+        tempIO.close()
```
rly, line 580: why does this need to be closed but not the above methods like `test_write_attributes_write_scalar_str` which also call
```
tempIO = ZarrIO(self.store, mode='w')
tempIO.open()
tempIO.write_attributes(...)
```

tests/unit/base_tests_zarrio.py
```diff
@@ -502,60 +567,79 @@ def test_write_attributes_write_3Darray_of_floats(self):
     def test_write_attributes_write_reference_to_datasetbuilder(self):
         data_1 = np.arange(100, 200, 10).reshape(2, 5)
         dataset_1 = DatasetBuilder('dataset_1', data_1)
-        testgroup = self.io._ZarrIO__file  # For testing we just use our file and create some attributes
+        tempIO = ZarrIO(self.store, mode='w')
+        tempIO.open()
+        testgroup = tempIO.file  # For testing we just use our file and create some attributes
         attr = {'attr1': dataset_1}
-        self.io.write_attributes(testgroup, attr)
+        tempIO = ZarrIO(self.store, mode='w')
```
rly: This confuses me. First, open the zarr store and get the file object. Then open it again and write some attributes. Why open it twice?

tests/unit/test_io_convert.py
rly: HDMF 3.5.0 was released so you can update the hdmf requirement and do this TODO in this PR or a new one

tests/unit/test_io_convert.py
oruebel: I changed the docs for the mixin classes to document the class members directly and also added a note to the module to clarify that the Mixins should always be used through inheritance.

tests/unit/test_io_convert.py
oruebel: Good catch. I updated the requirements and added the message here.

tests/unit/base_tests_zarrio.py
oruebel: In principle it is no longer necessary (as of the latest HDMF release) to call close here explicitly, because this is now taken care of by ``HDMFIO.__del__`` when the object is being garbage collected at the end of the function. However, in the test cases where we need to call ``open`` explicitly to test the low-level write functions, I think it is cleaner to then also explicitly call ``close``. I updated the tests to add missing ``close`` calls. Note, this is only for tests that use the low level write functions directly (e.g., to test ``write_dataset``). Most other tests can use the more standard pattern of using a context for the IO object so that we don't have to explicitly do the open and close.

tests/unit/base_tests_zarrio.py
oruebel: I don't think it is necessary. I updated the test case to avoid opening the file twice.

oruebel:
@rly thanks for the suggestions. I have now resolved all the items and this PR is now ready for you to take another look at for review. Specifically I:
* Removed the unused ``filepath`` parameter from the ``ZarrIO.get_builder_exists_on_disk``
* Increased the HDMF version to 3.5
* Fixed the documentation of class members of test mixin classes
* Consistently close IO objects in tests that we explicitly open 
* Updated the documentation to integrate new data stores

rly (approved):

rly:
Looks good to me. Thanks for the changes!

PR #72, opened 2023-05-08
Fix reading dtype from zarr dataset

alejoe91:
Hi guys, 

I'm experimenting now with writing to Zarr using a custom DataChunkIterator from [neuroconv](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/tools/spikeinterface/spikeinterfacerecordingdatachunkiterator.py#L8) bassed to the `ZarrDataIO`.

Everything works fine, but upon reading the file, I get that the `zarr_type` is not found in the attirbutes. However, this can be inferred from the `dtype` of the `zarr.Dataset`. 
Not sure this is the most elegant solution and I didn't have time to dig in and find a better one, but it works in my case :)

### DIFF
--- a/src/hdmf_zarr/backend.py
+++ b/src/hdmf_zarr/backend.py
@@ -1104,7 +1104,9 @@ class ZarrIO(HDMFIO):
             zarr_dtype = zarr_obj.attrs['zarr_dtype']
         elif hasattr(zarr_obj, 'dtype'):   # Fallback for invalid files that are mssing zarr_type
             zarr_dtype = zarr_obj.dtype
-            warnings.warn("Inferred dtype from zarr type. Dataset missing zarr_dtype: " + str(name) + "   " + str(zarr_obj)) 
+            warnings.warn(
+                "Inferred dtype from zarr type. Dataset missing zarr_dtype: " + str(name) + "   " + str(zarr_obj)
+            )
         else:
             raise ValueError("Dataset missing zarr_dtype: " + str(name) + "   " + str(zarr_obj))
 
--- a/src/hdmf_zarr/backend.py
+++ b/src/hdmf_zarr/backend.py
@@ -1100,7 +1100,7 @@ class ZarrIO(HDMFIO):
         if ret is not None:
             return ret
 
-        if 'zarr_dtype' not in zarr_obj.attrs:
+        if 'zarr_dtype' in zarr_obj.attrs:
             zarr_dtype = zarr_obj.attrs['zarr_dtype']
         elif hasattr(zarr_obj, 'dtype'):   # Fallback for invalid files that are mssing zarr_type
             zarr_dtype = zarr_obj.dtype
--- a/src/hdmf_zarr/backend.py
+++ b/src/hdmf_zarr/backend.py
@@ -670,6 +670,7 @@ class ZarrIO(HDMFIO):
                 io_settings['dtype'] = cls.__dtypes.get(io_settings['dtype'])
         try:
             dset = parent.create_dataset(name, **io_settings)
+            dset.attrs['zarr_dtype'] = io_settings['dtype']
         except Exception as exc:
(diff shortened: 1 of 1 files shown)


oruebel:
@alejoe91  it sounds like the `zarr_dtype` attribute is not being set when writing with DataChunkIterator. From a quick look I think the following code in `` __setup_chunked_dataset__`` 

https://github.com/hdmf-dev/hdmf-zarr/blob/e4e9543e53efaefc31b74c4fd8557353cb0b9121/src/hdmf_zarr/backend.py#L671-L675

is likely the problem and should be updated to set the `zarr_dtype` attribute as follows:

```
        type_str = self.__serial_dtype__(io_settings['dtype'])
        try:
            dset = parent.create_dataset(name, **io_settings)
        except Exception as exc:
            raise Exception("Could not create dataset %s in %s" % (name, parent.name)) from exc
        return dset
```

The fix you propose helps work around the issue if an invalid file is encountered but doesn't fix the issue of creating bad files.

alejoe91:
Yeah I figured there was probably a better solution! Should I update the PR with the fix you suggested @oruebel ?

src/hdmf_zarr/backend.py
```diff
@@ -1099,11 +1099,16 @@ def __read_dataset(self, zarr_obj, name):
...
 
-        if 'zarr_dtype' not in zarr_obj.attrs:
-            raise ValueError("Dataset missing zarr_dtype: " + str(name) + "   " + str(zarr_obj))
+        if hasattr(zarr_obj, 'dtype'):
+            zarr_dtype = zarr_obj.dtype
+        else:
+            if 'zarr_dtype' in zarr_obj.attrs:
+                zarr_dtype = zarr_obj.attrs['zarr_dtype']
+            else:
+                raise ValueError("Dataset missing zarr_dtype: " + str(name) + "   " + str(zarr_obj))
```
oruebel: I would suggest having the case where we use `zarr_dtype` attribute first and using the `zarr_obj.dtype` as a fallback, instead of the other way around. Also, I would suggest raising a warning when the dtype is inferred, just in case this should lead to any downstream errors to ease debugging and to make sure the user is still notified that the file has a problem.

```suggestion
        if 'zarr_dtype' not in zarr_obj.attrs:
            zarr_dtype = zarr_obj.attrs['zarr_dtype']   
        elif hasattr(zarr_obj, 'dtype'):   # Fallback for invalid files that are mssing zarr_type
            zarr_dtype = zarr_obj.dtype
            warnings.warn("Inferred dtype from zarr type. Dataset missing zarr_dtype: " + str(name) + "   " + str(zarr_obj)) 
        else:     
            raise ValueError("Dataset missing zarr_dtype: " + str(name) + "   " + str(zarr_obj))
```

oruebel:
> Should I update the PR with the fix you suggested

Yes please. I think we can keep your fix as well as since it fixes potential issues on read, but it would be best to make sure we write the file correctly.

src/hdmf_zarr/backend.py
alejoe91: done

alejoe91:
> > Should I update the PR with the fix you suggested
> 
> Yes please. I think we can keep your fix as well as since it fixes potential issues on read, but it would be best to make sure we write the file correctly.

Should be done ;)

alejoe91:
@oruebel 

fixed flake8. I don't think failing tests are due to this change

oruebel:
> I don't think failing tests are due to this change

I'll take a look at fixing tests tomorrow. One issue is that codecov has been pulled from PyPi and it looks like there are some other issues in the gallery tests. I'll work on that to make sure we can get all tests passing.

oruebel (approved):
Looks good. 

I created #80 to make sure we add roundtrip tests to cover this case and the changes from this PR

PR #78, opened 2023-05-09
Gallery Bug

mavaylon1:
Address 
```
DeprecationWarning: Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
```

### DIFF
--- a/requirements-min.txt
+++ b/requirements-min.txt
@@ -1,4 +1,4 @@
-hdmf==3.5.0
+hdmf==3.5.2
 zarr==2.11.0
 numcodecs==0.9.1
 pynwb==2.0.0
--- a/requirements-min.txt
+++ b/requirements-min.txt
@@ -3,4 +3,3 @@ zarr==2.11.0
 numcodecs==0.9.1
 pynwb==2.0.0
 setuptools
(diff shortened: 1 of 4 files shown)


mavaylon1:
@oruebel The changes I made solves the depreciation issue but introduces 
```
Exception: Could not create dataset first_name in / with shape (2,), dtype <class 'numpy.object_'>, and iosettings {}. Object dtype dtype('O') has no native HDF5 equivalent
```

mavaylon1:
@oruebel the fix seems to be to just update hdmf version to at least 3.5.2, which has the fix in hdmf for gallery deprecation, but below the newest 3.5.5

oruebel (approved):

oruebel:
@alejoe91 just FYI. This PR temporarily restricts the HDMF version, because 3.5.5 triggered errors in the tests in hdmf-zarr related to export from Zarr to HDF5. We are investigating the issue to create a proper fix. Restricting the HDMF version temporally helps to keep the CI running for now so we can continue working on other PRs while we fix the issue. Since this is related to conversion from Zarr to HDF5 this hopefully won't affect your work.

PR #87, opened 2023-05-15
Add Support for Python 3.11

mavaylon1:


mavaylon1:
@rly I'll continue to look into this tomorrow, but let me know if you have any insight. This seems to be an issue with numpy and numcodecs.

tox.ini
```diff
@@ -166,8 +170,8 @@ deps =
 commands = {[testenv:gallery]commands}
 
 # Test with python 3.10; pinned dev, doc, and optional reqs; pre-release run reqs
```
rly:
```suggestion
# Test with python 3.11; pinned dev, doc, and optional reqs; pre-release run reqs
```

rly:
You probably need to pin the latest versions of numpy and numcodecs in `requirements.txt`. Same for hdmf and pynwb. Only the latest versions of hdmf and pynwb allow you to install the latest version of numpy.

mavaylon1:
@rly I am not sure where the linux-python3.10-upgraded  is considering that it was replaced with  linux-python3.10-upgraded in the workflows
Otherwise this should be good to go for python 3.11

mavaylon1:
> @rly I am not sure where the linux-python3.10-upgraded is considering that it was replaced with linux-python3.10-upgraded in the workflows Otherwise this should be good to go for python 3.11

Ah it was branch protection. Solved

oruebel (approved):

PR #105, opened 2023-07-12
Update backend.py

mavaylon1:
- Update I/O for Zarr to support the linkage of ExternalResources (bringing this inline with hdmf)
- Remove python 3.7 support

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 # pinned dependencies to reproduce an entire development environment to use HDMF-ZARR
 hdmf==3.7.0
 zarr==2.11.0
-pynwb==2.3.0
+pynwb==2.3.2
 numpy==1.23.5
 numcodecs==0.11.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 # pinned dependencies to reproduce an entire development environment to use HDMF-ZARR
-hdmf==3.5.2
+hdmf==3.7.0
 zarr==2.11.0
 pynwb==2.3.0
 numpy==1.23.5
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 # pinned dependencies to reproduce an entire development environment to use HDMF-ZARR
-hdmf==3.7.0
+hdmf==3.9.0
 zarr==2.11.0
-pynwb==2.3.2
-numpy==1.23.5
+pynwb==2.5.0
+numpy==1.24
 numcodecs==0.11.0
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -62,7 +62,6 @@ body:
     attributes:
       label: Python Version
       options:
-        - "3.7"
         - "3.8"
         - "3.9"
         - "3.10"
--- a/.github/PULL_REQUEST_TEMPLATE/release.md
+++ b/.github/PULL_REQUEST_TEMPLATE/release.md
@@ -1,4 +1,4 @@
-Prepare for release of HDMF [version]
+Prepare for release of HDMF-Zarr [version]
 
 ### Before merging:
 - [ ] Major and minor releases: Update package versions in `requirements.txt`, `requirements-dev.txt`,
@@ -21,4 +21,4 @@ Prepare for release of HDMF [version]
    [GitHub releases page](https://github.com/hdmf-dev/hdmf-zarr/releases) with the changelog
 3. Check that the readthedocs "latest" and "stable" builds run and succeed
 4. Update [conda-forge/hdmf_zarr-feedstock](https://github.com/conda-forge/hdmf_zarr-feedstock) with the latest version number
-   and SHA256 retrieved from PyPI > HDMF > Download Files > View hashes for the `.tar.gz` file. Re-render as needed
+   and SHA256 retrieved from PyPI > HDMF-Zarr > Download Files > View hashes for the `.tar.gz` file. Re-render as needed
--- a/.github/workflows/deploy_release.yml
+++ b/.github/workflows/deploy_release.yml
@@ -28,11 +28,11 @@ jobs:
 
       - name: Run tox tests
         run: |
-          tox -e py310-upgraded
+          tox -e py311-upgraded
 
       - name: Build wheel and source distribution
         run: |
-          tox -e build-py310-upgraded
+          tox -e build-py311-upgraded
           ls -1 dist
 
       - name: Test installation from a wheel
--- a/.github/workflows/run_all_tests.yml
+++ b/.github/workflows/run_all_tests.yml
@@ -22,24 +22,21 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - { name: linux-python3.7-minimum      , test-tox-env: py37-minimum    , build-tox-env: build-py37-minimum    , python-ver: "3.7" , os: ubuntu-latest }
-          - { name: linux-python3.8              , test-tox-env: py38            , build-tox-env: build-py38            , python-ver: "3.8" , os: ubuntu-latest }
+          - { name: linux-python3.8-minimum      , test-tox-env: py38-minimum    , build-tox-env: build-py38-minimum    , python-ver: "3.8" , os: ubuntu-latest }
           - { name: linux-python3.9              , test-tox-env: py39            , build-tox-env: build-py39            , python-ver: "3.9" , os: ubuntu-latest }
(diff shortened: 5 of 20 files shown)


mavaylon1:
Fix #106

mavaylon1:
@oruebel After the updates, the gallery is failing with 
```
Exception: could not resolve dtype for VectorIndex 'spike_times_index'
```
Should we make the fix for that separate on its own ticket? I also believe this is linked to the external links workflow failing.
Other than that it seems bringing hdmf-zarr to hdmf 3.7 works.

oruebel:
> Exception: could not resolve dtype for VectorIndex 'spike_times_index'

That looks like another error due to https://github.com/hdmf-dev/hdmf/pull/848 triggering the rebuild of Containers such that we have to determine the data type rather than remembering the existing type. I suspect that this will most likely require a fix in the `ObjectMapper.convert_dtype` function in HDMF and a corresponding bug fix release for HDMF. I.e., I would suggest we do that first instead of releasing an update here with a known bug.

mavaylon1:
> > Exception: could not resolve dtype for VectorIndex 'spike_times_index'
> 
> That looks like another error due to [hdmf-dev/hdmf#848](https://github.com/hdmf-dev/hdmf/pull/848) triggering the rebuild of Containers such that we have to determine the data type rather than remembering the existing type. I suspect that this will most likely require a fix in the `ObjectMapper.convert_dtype` function in HDMF and a corresponding bug fix release for HDMF. I.e., I would suggest we do that first instead of releasing an update here with a known bug.

Right that makes sense, but I was suggesting we merge this and handle it separately before a release. @oruebel

oruebel:
> I was suggesting we merge this and handle it separately before a release

Let me check if I can figure this one out tomorrow so we can at least get an estimate of how hard it will be to fix this bug.

mavaylon1:
> > I was suggesting we merge this and handle it separately before a release
> 
> Let me check if I can figure this one out tomorrow so we can at least get an estimate of how hard it will be to fix this bug.

@oruebel thoughts?

requirements.txt
```diff
@@ -1,5 +1,5 @@
 # pinned dependencies to reproduce an entire development environment to use HDMF-ZARR
-hdmf==3.5.4
+hdmf==3.7.0
```
CodyCBakerPhD: Hey guys, is there a reason why you use exact version pinning here? It's causing problems in https://github.com/NeurodataWithoutBorders/nwbinspector/pull/388 and could cause problems elsewhere as well

Would there be catastrophic issues caused by looser versioning? e.g., `hdmf>=3.5.4`

CodyCBakerPhD: @rly @oruebel @mavaylon1

mavaylon1: It's to allow HDMF_ZARR to use the newest external resources, but I really don't have time this week to fix anything backend breaking until this week, which is too close to the dev/user days. I believe @oruebel was going to look into it last week.

mavaylon1: I will start at the end of the week but I will make this a project for the dev days.

oruebel:
When running `plot_convert_nwb_hdf5.py` from the gallery with the `test_gallery.py` script I get a different error then when running from the script from the command-line. I'm wondering whether the either the `test_gallery.py` is picking up an older version of HDMF or whether something has changed on the `dev` branch of HDMF.

mavaylon1:
1. python plot_convert_nwb_hdf5.py (w/ clear cache)
Calls export (goes through a bunch of super() calls to inherited export) > the export in io.py calls "manager.build" > which goes to type_map build > which goes to obj_mapper build >

Now upon initial inspection of the traceback
```
  File "/Users/mavaylon/Research/NWB/hdmf2/hdmf/src/hdmf/backends/hdf5/h5tools.py", line 1468, in __list_fill__
    raise e
  File "/Users/mavaylon/Research/NWB/hdmf2/hdmf/src/hdmf/backends/hdf5/h5tools.py", line 1465, in __list_fill__
    dset[:] = data
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/Users/mavaylon/opt/anaconda3/envs/NWB2/lib/python3.10/site-packages/h5py/_hl/dataset.py", line 1009, in __setitem__
    self.id.write(mspace, fspace, val, mtype, dxpl=self._dxpl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5d.pyx", line 280, in h5py.h5d.DatasetID.write
  File "h5py/_proxy.pyx", line 145, in h5py._proxy.dset_rw
  File "h5py/_conv.pyx", line 444, in h5py._conv.str2vlen
  File "h5py/_conv.pyx", line 95, in h5py._conv.generic_converter
  File "h5py/_conv.pyx", line 249, in h5py._conv.conv_str2vlen
TypeError: Can't implicitly convert non-string objects to strings
```
When you add a breakpoint in line 1467 of h5tools.py, you are confronted with an np.array of group builders. These builders are for the electrode group ADunit_32. The fact that it's a np.array is not correct (via the advice of Ryan) instead of a dataset of references. 

Update/Edit: Looking at the same set of builders, it is a np.array of Builders either way and so I don't think the format is wrong. You can verify by the original breakpoint at 1467 on h5tools and then by removing the cache and setting a breakpoint  at 1209 on h5tools and looking at 'data'. Both are np.arrays of Builders. What's even funnier is the _dtype for dset is "dtype('O')" (type n a few times), which to me is weird since that's what were trying to avoid (hence Oliver's change).

Now 1467 is within "list_fill" which is called by "write_dataset" (still within h5tools). The part of write_dataset you'd think we'd be in would be line 1265 to write an array of object references, BUT we are in line 1301. The options['dtype'] returned is utf8, which remind me of Oliver's addition of 
```
elif np.issubdtype(value.dtype, np.dtype('O')):
                    # Only variable-length strings should ever appear as generic objects.
                    # Everything else should have a well-defined type
                    ret_dtype = 'utf8'
```
This options['dtype'] shouldn't be utf8 from my understanding.

Now what about of we removed the clear cache. What would the options['dtype'] be? It is object. (Add a breakpoint at 1208 and see)

So they are different and we need to figure out what is setting the dtype.
Going back to Oliver's change, why was it added in the first place?

mavaylon1:
2. Exploring more into where 'dtype' is set, it seems to be set on line 744 on obj_mapper, which is in build.
So to connect back to the traceback flow where it leaves with calling build from obj_mapper. However, we don't seem to even enter the conditional that sets dtype='object'. 

Update:
It goes through line 747 on obj_map where it calls convert_dtype which is where Oliver's code sets it to utf8. (with cache)

mavaylon1:
3. I ran 
```
import logging

logger = logging.getLogger()
logger.setLevel(logging.DEBUG)

ch = logging.FileHandler('test.log', mode='w')
ch.setLevel(logging.DEBUG)
formatter = logging.Formatter('%(name)s - %(levelname)s - %(message)s')
ch.setFormatter(formatter)
logger.addHandler(ch)
```
I compared the two log files (with and without the cache) and there differences (one makes a new root while the non-cache reuses). I will take a look more tomorrow, but any insight would be great. @oruebel

oruebel:
> Going back to Oliver's change, why was it added in the first place

On export when the cache is being cleared, it means that HDMF is now constructing a new set of builders. On the "original" builders that were read from file (e.g., HDF5) the dtype is already set and so when we export the builders we don't need to determine the dtype. However, when the cache is being cleared and the builders are being rebuilt, then that means that we now need to determine the dtype from the array (e.g, the `h5py.dataset`). I.e., when we clear the cache the code takes a different path because it now needs to determine the dtypes to use for all the datasets, rather than being able to use the original dtypes. @rly why is the cache being cleared again for export? 

> So they are different and we need to figure out what is setting the dtype.

Correct, I believe the issue is that when export added to clear the cache it now means we are hitting corner cases in the logic to determine dtypes in the ObjectMapping that we did not encounter before. If I understand it correctly, you are seeing that the dtype is set to the wrong type by the code I added to handle the case where we have an object-type numpy array, i.e., it seems like we either: 1) need some logic here to be able to distinguish between an numpy array of references vs. a numpy array of strings or 2) need to figure out how to keep information about the original dtypes that are being used around so that we don't have to guess the dtype (e.g., either be removing the clear cache or having an option to rebuild builders but keeping dtype information from the previous builders).

rly:
@mavaylon1 Now that removing python 3.7 testing was done in a separate PR, is this PR still necessary? I merged dev into here and the changes do not look relevant anymore.

PR #111, opened 2023-07-30
Add parallel zarr

### LINKED ISSUE #101: [Feature]: Parallel Write Support for HDMF-Zarr

Parallel Write Support for HDMF-Zarr

Allow NWB files written using the Zarr backend to leverage multiple threads or CPUs to enhance speed of operation.
Objectives

Zarr is built to support efficient Python parallelization strategies, both multi-processing and multi-threaded
HDMF-Zarr currently handles all write operations (including buffering and slicing) without exposing the necessary controls to enable these strategies

Approach and Plan
Identify the best injection point for parallelization parameters in the io.write() stack of HDMF-Zarr
Progress and Next Steps
TODO
Background and References

_No response_

Identify the best injection point for parallelization parameters in the `io.write()` stack of HDMF-Zarr (as controlled via the `NWBZarrIO`)

Essentially revolving around the line https://github.com/hdmf-dev/hdmf/blob/2f9ec567ebe1df9fccb05f139d2f669661e50018/src/hdmf/backends/hdf5/h5_utils.py#L61 from the main repo (which might be what is used to delegate the command here as well?

Yes.

### Code of Conduct

CodyCBakerPhD:
fix hdmf-dev/hdmf-zarr#101 

## Motivation

Zarr supports efficient parallelization, but enabling it seamlessly with only a single argument (`number_of_jobs` at `io.write`) took a bit of effort.

Currently seeing progressive speedups with the attached dummy script as the number of jobs increases; on the DANDI Hub ~160s for 1 CPU, . Will make an averaged plot over the number of jobs to use for reference

Screenshot 2023-07-30 at 5 37 08 PM

Screenshot 2023-07-30 at 5 37 30 PM

Screenshot 2023-07-30 at 5 38 08 PM

Will make a full averaged plot over the number of jobs to use for reference

Opening in draft while I assess what all is still necessary and what can still be optimized in terms of worker/job initialization

Also will have to think on how to add tests; I suppose just adding some that use 2 jobs and making sure it works should be enough

## How to test the behavior?

```
import numpy as np
from pathlib import Path

from pynwb.testing.mock.file import mock_NWBFile
from pynwb.testing.mock.base import mock_TimeSeries
from neuroconv.tools.hdmf import SliceableDataChunkIterator

number_of_jobs = 1  # increase according to screenshot

dat_file_path = "/home/jovyan/performance_tests/example_data.dat"
n_frames = 30000 * 60 * 2
n_channels = 384
data_shape = (n_frames, n_channels)
dtype = "int16"
memory_map = np.memmap(filename=dat_file_path, dtype=dtype, mode="r", shape=data_shape)  # about ~2.75 GB of data

nwbfile = mock_NWBFile()
time_series = mock_TimeSeries(data=SliceableDataChunkIterator(data=memory_map, buffer_gb=0.1 / number_of_jobs))
nwbfile.add_acquisition(time_series)

zarr_top_level_path = f"/home/jovyan/Downloads/example_parallel_zarr_{number_of_jobs}.nwb"
with NWBZarrIO(path=zarr_top_level_path, mode="w") as io:
    io.write(nwbfile, number_of_jobs=number_of_jobs)
```

## Checklist


oruebel:
> I suppose just adding some that use 2 jobs and making sure it works should be enough

Focusing on "functions correctly" in the unit tests is fine. Doing tests to check for performance is tricky.

> enabling it seamlessly with only a single argument (`number_of_jobs` at `io.write`) took a bit of effort.

Thanks for the hard work to make this easy for users. 

It would be great to also add a short gallery tutorial for parallel write here https://github.com/hdmf-dev/hdmf-zarr/tree/dev/docs/gallery

A nice-to-have feature would be to enhance ``ZarrIO.export`` (https://github.com/hdmf-dev/hdmf-zarr/blob/d47fc8249e76a371fff990b07218a088e59ea590/src/hdmf_zarr/backend.py#L231) to use parallel write when converting from HDF5 to Zarr. One way to achieve this would be to automatically wrap datasets in DataChunkIterator in [``ZarrIO.write_dataset``](https://github.com/hdmf-dev/hdmf-zarr/blob/d47fc8249e76a371fff990b07218a088e59ea590/src/hdmf_zarr/backend.py#L697). This is probably beyond this PR and can be addressed later in a separate issue. Just wanted to mention it.

bendichter:
Good idea, @oruebel I'll create a follow-up issue for that

tests/unit/test_parallel_write.py
```diff
@@ -0,0 +1,57 @@
...
+        return iterator
+
+def test_parallel_write(tmpdir):
+    number_of_jobs = 2
+    column = VectorData(name="TestColumn", description="", data=TestDataChunkIterator(data=np.array([1., 2., 3.])))
+    dynamic_table = DynamicTable(name="TestTable", description="", columns=[column])
+
+    zarr_top_level_path = str(tmpdir / f"example_parallel_zarr_{number_of_jobs}.zarr")
+    with ZarrIO(path=zarr_top_level_path, mode="w") as io:
+        io.write(dynamic_table, number_of_jobs=number_of_jobs)
```
CodyCBakerPhD: First attempt at just writing a DynamicTable object alone gives the following error stack

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[9], line 56
     54 zarr_top_level_path = str(tmpdir / f"example_parallel_zarr_{number_of_jobs}.zarr")
     55 with ZarrIO(path=zarr_top_level_path, mode="w") as io:
---> 56     io.write(dynamic_table, number_of_jobs=number_of_jobs)

File ~/GitHub/hdmf/src/hdmf/utils.py:644, in docval.<locals>.dec.<locals>.func_call(*args, **kwargs)
    642 def func_call(*args, **kwargs):
    643     pargs = _check_args(args, kwargs)
--> 644     return func(args[0], **pargs)

File ~/GitHub/hdmf-zarr/src/hdmf_zarr/backend.py:231, in ZarrIO.write(self, **kwargs)
    229 """Overwrite the write method to add support for caching the specification"""
    230 cache_spec = popargs('cache_spec', kwargs)
--> 231 super(ZarrIO, self).write(**kwargs)
    232 if cache_spec:
    233     self.__cache_spec()

File ~/GitHub/hdmf/src/hdmf/utils.py:644, in docval.<locals>.dec.<locals>.func_call(*args, **kwargs)
    642 def func_call(*args, **kwargs):
    643     pargs = _check_args(args, kwargs)
--> 644     return func(args[0], **pargs)

File ~/GitHub/hdmf/src/hdmf/backends/io.py:81, in HDMFIO.write(self, **kwargs)
     79 """Write a container to the IO source."""
     80 container = popargs('container', kwargs)
---> 81 f_builder = self.__manager.build(container, source=self.__source, root=True)
     82 self.write_builder(f_builder, **kwargs)
     84 if self.external_resources_path is not None:

File ~/GitHub/hdmf/src/hdmf/utils.py:644, in docval.<locals>.dec.<locals>.func_call(*args, **kwargs)
    642 def func_call(*args, **kwargs):
    643     pargs = _check_args(args, kwargs)
--> 644     return func(args[0], **pargs)

File ~/GitHub/hdmf/src/hdmf/build/manager.py:171, in BuildManager.build(self, **kwargs)
    167             raise ValueError("Cannot change container_source once set: '%s' %s.%s"
    168                              % (container.name, container.__class__.__module__,
    169                                 container.__class__.__name__))
    170 # NOTE: if exporting, then existing cached builder will be ignored and overridden with new build result
--> 171 result = self.__type_map.build(container, self, source=source, spec_ext=spec_ext, export=export)
    172 self.prebuilt(container, result)
    173 self.__active_prebuilt(result)

File ~/GitHub/hdmf/src/hdmf/utils.py:644, in docval.<locals>.dec.<locals>.func_call(*args, **kwargs)
    642 def func_call(*args, **kwargs):
    643     pargs = _check_args(args, kwargs)
--> 644     return func(args[0], **pargs)

File ~/GitHub/hdmf/src/hdmf/build/manager.py:763, in TypeMap.build(self, **kwargs)
    760 source, spec_ext, export = getargs('source', 'spec_ext', 'export', kwargs)
    762 # get the ObjectMapper to map between Spec objects and AbstractContainer attributes
--> 763 obj_mapper = self.get_map(container)
    764 if obj_mapper is None:
    765     raise ValueError('No ObjectMapper found for container of type %s' % str(container.__class__.__name__))

File ~/GitHub/hdmf/src/hdmf/utils.py:644, in docval.<locals>.dec.<locals>.func_call(*args, **kwargs)
    642 def func_call(*args, **kwargs):
    643     pargs = _check_args(args, kwargs)
--> 644     return func(args[0], **pargs)

File ~/GitHub/hdmf/src/hdmf/build/manager.py:705, in TypeMap.get_map(self, **kwargs)
    703     namespace, data_type = self.get_container_cls_dt(container_cls)
    704     if namespace is None:
--> 705         raise ValueError("class %s is not mapped to a data_type" % container_cls)
    706 else:
    707     data_type = self.get_builder_dt(obj)

ValueError: class <class 'hdmf.common.table.DynamicTable'> is not mapped to a data_type
```

CodyCBakerPhD: Fixed by including `manager=get_manager()`

oruebel:
BTW, if pickling is an issue https://github.com/cloudpipe/cloudpickle may be useful

CodyCBakerPhD:
> BTW, if pickling is an issue https://github.com/cloudpipe/cloudpickle may be useful

Yeah I looked into this after the dev days; `cloudpickle` definitely seems useful in general, I've seen cases before of custom `.pkl` files that can't be loaded in environments different from what they were saved in, which is what that package seems to emphasize. And the general approach of pickling by value definitely seems preferable than by reference

However the 'problem' that is more relevant to us in this context is dealing with class types that _do_ have picking methods defined (strictly speaking) but ones that are more inefficient than they need to be. In particular, numpy array/memmap and zarr datasets serialize their data contents by default instead of passing minimal information (store or file paths) needed to reinstantiate the the object in another workers memory

It's also not really even a 'problem' here strictly speaking, since this PR only assumes the iterator has such methods defined and builds around it - it's really the burden of the creator of such iterators (mainly NeuroConv https://github.com/catalystneuro/neuroconv/pull/536) to make them efficient

tests/unit/test_parallel_write.py
```diff
@@ -0,0 +1,185 @@
...
+
+    zarr_top_level_path = str(tmpdir / "test_mixed_iterator_pickleability.zarr")
+    with ZarrIO(path=zarr_top_level_path,  manager=get_manager(), mode="w") as io:
+        io.write(dynamic_table, number_of_jobs=number_of_jobs)
+
+    # TODO: roundtrip currently fails due to read error
+    #with ZarrIO(path=zarr_top_level_path, mode="r") as io:
+    #    dynamic_table_roundtrip = io.read()
+    #    data_roundtrip = dynamic_table_roundtrip["TestColumn"].data
+    #    assert_array_equal(data_roundtrip, data)
```
CodyCBakerPhD: see #116 for blocking issue including this extra layer to the tests here

tests/unit/test_parallel_write.py
```diff
@@ -0,0 +1,185 @@
...
+        io.write(dynamic_table, number_of_jobs=number_of_jobs)
+
+    tqdm_out_value = tqdm_out.getvalue()
+    assert expected_desc_pickleable in tqdm_out_value
+    assert expected_desc_not_pickleable in tqdm_out_value
+
+
+def test_extra_args(tmpdir):
+    pass # TODO? Should we test if the other arguments like thread count can be passed?
+    # I mean, anything _can_ be passed due to dynamic **kwargs, but how to test if it was actually used? Seems difficult...
```
CodyCBakerPhD, line 232: @oruebel thoughts on if tests for this are needed, or strategy to take?

For the `max_threads_per_process` and `multiprocessing_context` arguments that are usually encouraged to just default, but control is left to allow them to be overridden for architectures that utilize it

tests/unit/test_parallel_write.py
```diff
@@ -0,0 +1,185 @@
...
+from numpy.testing import assert_array_equal
+from hdmf_zarr import ZarrIO
+from hdmf.common import DynamicTable, VectorData, get_manager
+from hdmf.data_utils import GenericDataChunkIterator, DataChunkIterator
+
+try:
+    import tqdm  # noqa: F401
+    TQDM_INSTALLED = True
+except ImportError:
+    TQDM_INSTALLED = False
```
CodyCBakerPhD, line 18: HDMF has https://github.com/hdmf-dev/hdmf/blob/dev/requirements-opt.txt

but I see no such analog here - how should I proceed on adding an optional dependency such as `tqdm`?

CodyCBakerPhD:
OK the tests have more or less shaped together now

Let me know if there's more to add

tests/unit/test_parallel_write.py
oruebel: @rly thoughts. I think adding a similar `requirements-opt.txt` is fine.

CodyCBakerPhD: @rly Ping

tests/unit/test_parallel_write.py
oruebel: I added some thoughts on issue #116. Long story short, I think if you modify line 115 to set the `id` parameter correctly, then this should work (I think).

tests/unit/test_parallel_write.py
```diff
@@ -0,0 +1,185 @@
...
+    #with ZarrIO(path=zarr_top_level_path, mode="r") as io:
+    #    dynamic_table_roundtrip = io.read()
+    #    data_roundtrip = dynamic_table_roundtrip["TestColumn"].data
+    #    assert_array_equal(data_roundtrip, data)
+
+def test_mixed_iterator_pickleability(tmpdir):
+    number_of_jobs = 2
+    pickleable_column = VectorData(name="TestGenericColumn", description="", data=PickleableDataChunkIterator(data=np.array([1., 2., 3.])))
+    not_pickleable_column = VectorData(name="TestClassicColumn", description="", data=NotPickleableDataChunkIterator(data=np.array([4., 5., 6.])))
+    dynamic_table = DynamicTable(name="TestTable", description="", columns=[pickleable_column, not_pickleable_column])
```
oruebel:
```suggestion
    dynamic_table = DynamicTable(name="TestTable", description="", columns=[pickleable_column, not_pickleable_column], id=np.array([1., 2., 3.]))
```

tests/unit/test_parallel_write.py
```diff
@@ -0,0 +1,185 @@
...
+    def _get_data(self, selection: Tuple[slice]) -> np.ndarray:
+        return self.data[selection]
+    
+def test_parallel_write(tmpdir):
+    number_of_jobs = 2
+    data = np.array([1., 2., 3.])
+    column = VectorData(name="TestColumn", description="", data=PickleableDataChunkIterator(data=data))
+    dynamic_table = DynamicTable(name="TestTable", description="", columns=[column])
+
+    zarr_top_level_path = str(tmpdir / "test_parallel_write.zarr")
```
oruebel, line 83: I think using `os.path.join(tmpdir, "test_parallel_write.zarr")` is safer here. (Similar also in the other test cases)

src/hdmf_zarr/utils.py
```diff
@@ -1,22 +1,129 @@
...
+# Inherited from SpikeInterface experience
+
+# see
+# https://stackoverflow.com/questions/10117073/how-to-use-initializer-to-set-up-my-multiprocess-pool
+# the tricks is : theses 2 variables are global per worker
+# so they are not share in the same process
+global _worker_context
+global _operation_to_run
+
+
```
oruebel, line 39: As far as I can tell the module level functions `_write_buffer_zarr`,  `_is_pickleable`, `function_wrapper`, and `initializer_wrapper` are actually specific for `ZarrIODataChunkIteratorQueue` (i.e., they are only used there). If that is the case, I would suggest declaring them as static methods as part of the `ZarrIODataChunkIteratorQueue`  class rather than at the module level (which implies more general functionality that is not specific to a particular class).

tests/unit/test_parallel_write.py
oruebel: Let's just mirror the `requirements-opt.txt` approach from HDMF.

tests/unit/test_parallel_write.py
oruebel: Yes, testing those particular parameters is tricky when they only being passed in function calls during `ZarrIO.write` where you don't have an easy way to inspect those values. The question is, does `max_threads_per_process` need to be set separately for every call to `exhaust_queue` (i.e., does this change between calls) or should this really be a property of the `ZarrIODataChunkIteratorQueue` instead? If the latter, then you could check after you call `ZarrIO.write` whether `ZarrIO.__dci_queue.max_threads_per_process` has the correct setting. 

Otherwise, I would add empty tests for the cases that you would like to check and skip those checks and then add a comment why this is not being tested.

oruebel:
@CodyCBakerPhD sorry for the delay in responding to your questions. I hope the comment I added help clarify the questions you had raised. Please ping me in case I missed anything.

tests/unit/test_parallel_write.py
CodyCBakerPhD: What do you mean by 'safer'?

I usually prefer `pathlib` for the better readability (one must know implicitly that the ordered `args` of `os.path.join` represent subfolder locations to be parsed by system-specific folder separators; whereas `/` operator on `pathlib.Path` objects is the explicit way of both indicating it in code and resolving it based on platform)

tests/unit/test_parallel_write.py
CodyCBakerPhD: Right you are - working now!

tests/unit/test_parallel_write.py
CodyCBakerPhD: Fixed

tests/unit/test_parallel_write.py
CodyCBakerPhD: It seems as if the `ZarrIODataChunkIteratorQueue` [is only instantiated at time](https://github.com/hdmf-dev/hdmf-zarr/blob/dev/src/hdmf_zarr/backend.py#L118) of `ZarrIO` instantiation; whereas both practically and in terms of abstract design, each of these keyword arguments (the most important of which, `number_of_jobs`) are specific to the call to the method `io.write` (which then deploys to `write_group`/`write_builder`, etc.) as opposed to being properties of the Queue or by extension, the IO class itself

This is also to avoid confusion regarding other IO operations, such as `io.read` - if we had `ZarrIO(number_of_jobs=4, ...)` one might think any operation performed using that IO would be parallelized when in reality it would still only be the `write` call

I suppose I could still set them as instance attributes of the `ZarrIODataChunkIteratorQueue` class that are always `None` until the call to `exhaust_queue` with `number_of_jobs > 1` + whatever extra argument specifications, and then update the attributes to non-`None` values in that event, which would at least let us test their propagation to the right level

How does that sound?

tests/unit/test_parallel_write.py
CodyCBakerPhD: OK think I got them all, but there seems to be a fair number of places that need to be updated (CI calls; various hidden config files, etc) so someone should probably double-check

src/hdmf_zarr/utils.py
CodyCBakerPhD: Sounds fair

Fixed; all non-global definitions moved attached to the Queue class as staticmethods

tests/unit/test_parallel_write.py
oruebel: Thanks for the clarification. I didn't realize this is using `pathlib`. I misread it as just doing manual string manipulation. This if fine.

setup.py
```diff
@@ -26,6 +26,7 @@
     'numcodecs==0.11.0; python_version >= "3.8"',
     'pynwb>=2.3.2',
     'setuptools',
+    'threadpoolct>=3.2.0',
```
oruebel, line 29: It looks `threadpoolct` is a new requirement. Does `threadpoolct` also need to be added to one of the `requirements*.txt` files?

requirements-opt.txt
```diff
@@ -0,0 +1 @@
+tqdm==4.65.0
```
oruebel, line 1: @rly should optional requirements also be listed in `setup.py` or only strict requirements?

tests/unit/test_parallel_write.py
oruebel: Thanks. @rly could you review this PR and check that the requirements (in particular the new optional requirements) are correct and that we didn't miss anything.

tests/unit/test_parallel_write.py
oruebel: > whereas both practically and in terms of abstract design, each of these keyword arguments (the most important of which, `number_of_jobs`) are specific to the call to the method `io.write`

That makes sense. To take this line-of-thought another step further, I think `ZarrIODataChunkIteratorQueue`  as a whole is really specific to `io.write`. As such, an alternative approach could be to instead of in `ZarrIO.__init__` we could create a new `ZarrIODataChunkIteratorQueue` every time we call `io.write` and allow the user to optionally supply their own `ZarrIODataChunkIteratorQueue`. I.e., something like `io.write(dci_queue=ZarrIODataChunkIteratorQueue(number_of_jobs=4)`.  In this way, the user would supply settings like `number_of_jobs` (which are specific to `ZarrIODataChunkIteratorQueue`) explicitly to the queue, rather than indirectly via `io.write`.  What do you think?

oruebel (commented):
Overall, this looks good to me.

requirements-opt.txt
rly: Optional requirements can be set up in `setup.py` with the key:
```
    extras_require={
        "tqdm": ["tqdm>=4.41.0"],
    },
```

setup.py
rly: yes

rly, line 29:
```suggestion
    'threadpoolctl>=3.2.0',
```
Just to double check, this should be `threadpoolctl` right? `threadpoolct` is not on PyPI. I'm surprised this works as is...

tests/unit/test_parallel_write.py
rly: Sorry I missed these pings. I will review

oruebel:
@CodyCBakerPhD also, could you please add an entry for these changes to the `CHANGLOG.md`

src/hdmf_zarr/backend.py
```diff
@@ -837,7 +1009,7 @@ def write_dataset(self, **kwargs):  # noqa: C901
         # Exhaust the DataChunkIterator if the dataset was given this way. Note this is a no-op
         # if the self.__dci_queue is empty
         if exhaust_dci:
-            self.__dci_queue.exhaust_queue()
+            self.__dci_queue.exhaust_queue(number_of_jobs=number_of_jobs)
```
rly, line 1012: Should this use the 'max_threads_per_process' and 'multiprocessing_context' kwargs?

src/hdmf_zarr/backend.py
```diff
@@ -336,15 +463,17 @@ def write_group(self, **kwargs):
...
 
         datasets = builder.datasets
         if datasets:
             for dset_name, sub_builder in datasets.items():
                 self.write_dataset(parent=group,
                                    builder=sub_builder,
                                    link_data=link_data,
-                                   exhaust_dci=exhaust_dci)
+                                   exhaust_dci=exhaust_dci,
+                                   number_of_jobs=number_of_jobs)
```
rly, line 476: Should these pass the 'max_threads_per_process' and 'multiprocessing_context' kwargs?

src/hdmf_zarr/utils.py
```diff
@@ -28,12 +48,18 @@ def __init__(self):
...
         Internal helper function used to read a chunk from the given DataChunkIterator
         and write it to the given Dataset
         :param dset: The Dataset to write to
+        :type dset: HDMFDataset
         :param data: The DataChunkIterator to read from
+        :type data: DataChunkIterator
+        :param buffer_index: The index of the next chunk within the DataChunkIterator, optional.
+        Only used with multiprocessing for serializing the selections of the iteration.
+        Default is to perform direct iteration on the in-memory DataChunkIterator for the single process.
+        :type buffer_index: integer or None
```
rly: buffer_index is not listed in the method kwargs above

src/hdmf_zarr/utils.py
```diff
@@ -63,19 +89,115 @@ def __write_chunk__(cls, dset, data):
...
+                        from tqdm import tqdm
+
+                        results = tqdm(iterable=results, **progress_bar_options)
+                        
+                        # exector map must be iterated to deploy commands over jobs
+                        for size_in_MB, result in zip(size_in_MB_per_iteration, results):
+                            results.update(n=int(size_in_MB))  # int() to round down to nearest integer for better display
+                    except Exception as exception:  # Import warnings are also issued at the level of the iterator instantiation
+                        warn(f"Unable to setup progress bar due to\ntype(exception): str(exception)\n\n{traceback.format_exc()}")
+                        # exector map must be iterated to deploy commands over jobs
```
rly:
```suggestion
                        # executor map must be iterated to deploy commands over jobs
```

src/hdmf_zarr/utils.py
```diff
@@ -63,19 +89,115 @@ def __write_chunk__(cls, dset, data):
...
+                        for size_in_MB, result in zip(size_in_MB_per_iteration, results):
+                            results.update(n=int(size_in_MB))  # int() to round down to nearest integer for better display
+                    except Exception as exception:  # Import warnings are also issued at the level of the iterator instantiation
+                        warn(f"Unable to setup progress bar due to\ntype(exception): str(exception)\n\n{traceback.format_exc()}")
+                        # executor map must be iterated to deploy commands over jobs
+                        for result in results:
+                            pass
+                else:
+                    # exector map must be iterated to deploy commands over jobs
```
rly:
```suggestion
                    # executor map must be iterated to deploy commands over jobs
```

src/hdmf_zarr/utils.py
CodyCBakerPhD: Ah, that was leftover from early attempts. Passing the buffer selection explicitly turned out to be more efficient

CodyCBakerPhD: We also decided not to alter `__write_chunk__` at all in this case

rly: Got it. I'll remove that from the docstring

rly:
```suggestion
```

src/hdmf_zarr/backend.py
CodyCBakerPhD: I adjusted things towards @oruebel's recommendation on attributes of an instance of the Queue class

I wasn't quite sure if it would work so I tried it on a separate branch #118, though it appears to be working, so switching to that

It does end up simplifying a lot of these docvals and argument passing

src/hdmf_zarr/backend.py
CodyCBakerPhD: See above

tests/unit/test_parallel_write.py
CodyCBakerPhD: I think that makes more sense; I wasn't entirely sure if it would work so I tried it safely on a separate branch, moving this PR over to that

#118

CodyCBakerPhD:
replaced by #118

src/hdmf_zarr/backend.py
oruebel: > it appears to be working, so switching to that
> 
> It does end up simplifying a lot of these docvals and argument passing

Thanks for all your efforts on this and your patients with all the changes and discussions. I'm glad that our discussions are helping to make the PR better, but I do appreciate that this has taken a lot of effort on your end. But I think this is coming along nicely and that this should be ready to merge soon.

PR #115, opened 2023-08-19
numpy degrade for python 3.7

mavaylon1:
## Motivation

Update python that supports 3.7. This is meant to address the fact that the Read the Docs for HDMF-Zarr does not build.

## How to test the behavior?
```
```
Run the tests and check read the docs build
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Bug fixes
 * Fixed error in deploy workflow. @mavaylon1 [#109](https://github.com/hdmf-dev/hdmf-zarr/pull/109)
-* fixed build error for ReadtheDocs. @mavaylon1 [#115](https://github.com/hdmf-dev/hdmf-zarr/pull/115)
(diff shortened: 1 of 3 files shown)


rly:
As a quick fix, this is fine but we also do not need to support Python 3.7 anymore unless there is a good reason to. We have migrated HDMF to Python 3.8 minimum.

mavaylon1:
> As a quick fix, this is fine but we also do not need to support Python 3.7 anymore unless there is a good reason to. We have migrated HDMF to Python 3.8 minimum.

Correct and there is a PR to make that upgrade (#105). What made us stop was the fact the export/cache bug preventing us from using a more recent version of HDMF to then drop 3.7. @rly

mavaylon1:
@oruebel @rly This does fix this build issue. I will clean it up tomorrow. I am aware of 3.7 being dropped and the export/cache issue is the main bump we will get through to bring hdmf-zarr up to speed with hdmf.

oruebel (approved):

PR #118, opened 2023-08-31
Parallel zarr with Queue instance attributes

### LINKED ISSUE #101: [Feature]: Parallel Write Support for HDMF-Zarr

CodyCBakerPhD:
fix #101 

replace #111 

## Motivation

Zarr supports efficient parallelization, but enabling it seamlessly with only a single argument (`number_of_jobs` at `io.write`) took a bit of effort.

Currently seeing progressive speedups with the attached dummy script as the number of jobs increases; on the DANDI Hub ~160s for 1 CPU, . Will make an averaged plot over the number of jobs to use for reference

Screenshot 2023-07-30 at 5 37 08 PM

Screenshot 2023-07-30 at 5 37 30 PM

Screenshot 2023-07-30 at 5 38 08 PM

Will make a full averaged plot over the number of jobs to use for reference

Opening in draft while I assess what all is still necessary and what can still be optimized in terms of worker/job initialization

Also will have to think on how to add tests; I suppose just adding some that use 2 jobs and making sure it works should be enough

## How to test the behavior?

```
import numpy as np
from pathlib import Path

from pynwb.testing.mock.file import mock_NWBFile
from pynwb.testing.mock.base import mock_TimeSeries
from neuroconv.tools.hdmf import SliceableDataChunkIterator

number_of_jobs = 1  # increase according to screenshot

dat_file_path = "/home/jovyan/performance_tests/example_data.dat"
n_frames = 30000 * 60 * 2
n_channels = 384
data_shape = (n_frames, n_channels)
dtype = "int16"
memory_map = np.memmap(filename=dat_file_path, dtype=dtype, mode="r", shape=data_shape)  # about ~2.75 GB of data

nwbfile = mock_NWBFile()
time_series = mock_TimeSeries(data=SliceableDataChunkIterator(data=memory_map, buffer_gb=0.1 / number_of_jobs))
nwbfile.add_acquisition(time_series)

zarr_top_level_path = f"/home/jovyan/Downloads/example_parallel_zarr_{number_of_jobs}.nwb"
with NWBZarrIO(path=zarr_top_level_path, mode="w") as io:
    io.write(nwbfile, number_of_jobs=number_of_jobs)
```

## Checklist


src/hdmf_zarr/backend.py
```diff
@@ -189,17 +189,54 @@ def load_namespaces(cls, namespace_catalog, path, namespaces=None):
...
-        cache_spec = popargs('cache_spec', kwargs)
+        """Overwrite the write method to add support for caching the specification and parallelization."""
+        cache_spec, number_of_jobs, max_threads_per_process, multiprocessing_context = popargs(
+            "cache_spec", "number_of_jobs", "max_threads_per_process", "multiprocessing_context", kwargs
+        )
+
+        self.__dci_queue = ZarrIODataChunkIteratorQueue(
+            number_of_jobs=number_of_jobs,
+            max_threads_per_process=max_threads_per_process,
+            multiprocessing_context=multiprocessing_context,
```
oruebel, line 236: Should `self.__dci_queue` then also be set back again to `None` at the end of `write(...)` or is it still needed after write?

oruebel:
ReadTheDocs shows the following error due to the added `threadpoolctl` requirement.

```
ERROR: Could not find a version that satisfies the requirement threadpoolctl==3.2.0 (from versions: 1.0.0, 1.1.0, 2.0.0, 2.1.0, 2.2.0, 3.0.0, 3.1.0)
ERROR: No matching distribution found for threadpoolctl==3.2.0
```

Maybe the [`readthedocs.yaml`](https://github.com/hdmf-dev/hdmf-zarr/blob/dev/.readthedocs.yaml) or the `requirements.txt` file need some adjustment.

src/hdmf_zarr/backend.py
CodyCBakerPhD: If we do that, we can't test the attribute propagation; I also don't know what it would accomplish since it will (should) be empty by that point anyway

CodyCBakerPhD:
> Maybe the [readthedocs.yaml](https://github.com/hdmf-dev/hdmf-zarr/blob/dev/.readthedocs.yaml) or the requirements.txt file need some adjustment.

I do believe this is an issue with the version of Python being used to compile the docs - do you know what that is?

You can also specify it explicitly in the config file like here: https://github.com/catalystneuro/neuroconv/blob/main/.readthedocs.yaml#L10-L11

Alternative would I suppose be to lower the exact version pin for the CI, but I defer to how y'all prefer to have all that setup

tests/unit/test_parallel_write.py
```diff
@@ -0,0 +1,271 @@
...
+        with ZarrIO(path=zarr_top_level_path,  manager=get_manager(), mode="w") as io:
+            io.write(
+                container=dynamic_table,
+                number_of_jobs=number_of_jobs,
+                max_threads_per_process=test_max_threads_per_process,
+                multiprocessing_context=test_multiprocessing_context
+            )
+
+            assert io._ZarrIO__dci_queue.max_threads_per_process == test_max_threads_per_process
+            assert io._ZarrIO__dci_queue.multiprocessing_context == test_multiprocessing_context
```
CodyCBakerPhD, line 267: Out of curiosity, can anyone direct me to an explanation of _why_ the `ZarrIO.__dci_queue` attribute ends up as `io._ZarrIO__dci_queue` here?

src/hdmf_zarr/backend.py
oruebel: > I also don't know what it would accomplish

I was mainly thinking about keeping a "clean" state and avoiding potential future side-effects, but I don't think it's a big problem since only `ZarrIO.write` uses it and it resets it at the beginning. E.g., a potential side-effect would be for `ZarrIO.export`, but as far as I recall it is just call `ZarrIO.write` so I think this is fine as well. Long-story-short, I think it's fine to not set `self.__dci_queue` back to None since it is useful for the tests.

tests/unit/test_parallel_write.py
oruebel: That is just Python's way of supporting `private` member variables and functions by mangling names. This part of the Python docs explains this briefly https://docs.python.org/3/tutorial/classes.html#private-variables

CodyCBakerPhD: Thanks for the pointer

Learn something new every day lol

rly:
Some tests define a new `ZarrIO` and call `write_dataset` directly. Because `self.__dci_queue` is initialized only in `write` and now `export`, these tests fail because `self.dci_queue` is None. If these tests are meant to test `write_dataset` in a unit test fashion, then these tests need to be adjusted so that `write` is called first or the `__dci_queue` variable is otherwise set. If these tests are meant to be integration tests, then these tests need to be adjusted so that they call `write` instead of `write_dataset` which users would not be doing.

CodyCBakerPhD:
Ahh good catch: https://github.com/hdmf-dev/hdmf-zarr/blob/6c13e14927eea985d53174d8580224c97d65707a/src/hdmf_zarr/backend.py#L839-L840

Since the method is not marked as private I'll just instantiate a standard non-parallel Queue at that point then

oruebel:
> If these tests are meant to test `write_dataset` in a unit test fashion,

Those should be unit tests, since write_dataset is not a a function that a user should ever call, but is used internally. The tests should be adjusted to manually set the dci_queue variable before calling write_dataset. Alternatively, we could also add `if __dci_queue is None` at the beginning of write_dataset to set it if it is not initialized (which may be safer)

oruebel:
> Since the method is not marked as private I'll just instantiate a standard non-parallel Queue at that point then

Sorry, I didn't see your comment until after I posted my other response. I agree, "instantiate a standard non-parallel Queue at that point then" is the way to go.

CodyCBakerPhD:
Surprised that 3.7 is still supported here - is there a timeline for when that will be dropped?

CodyCBakerPhD:
Otherwise, the currently failing CI is, I believe, due to the version pin of `hdmf==3.5.4`, which this feature requires some changes available on `hdmf>=3.9.0`

oruebel:
> Surprised that 3.7 is still supported here - is there a timeline for when that will be dropped?

This is due to the pin on the HDMF version. Once we have the issue with references on export resolved we'll update the HDMF version and then we can also update the tests. @mavaylon1 is working on the issue.

oruebel:
> Otherwise, the currently failing CI is, I believe, due to the version pin of `hdmf==3.5.4`, which this feature requires some changes available on `hdmf>=3.9.0`

To get the tests to run for now, you could just increase the hdmf version on this PR so we can see that the CI is running. There will be a couple of tests that fail for export, but at least we can then see that everything is working in the CI. Aside from the bug on export, I believe you can safely use the current version of HDMF without having to change anything else in the code.

oruebel:
@CodyCBakerPhD with #120 now merged, the dev branch now supports the latest HDMF. Could you sync this PR with the dev branch to see if that fixes the failing tests now so that we can move forward with merging this PR as well.

CodyCBakerPhD:
@oruebel Done, not sure what's up with the coverage workflows though

oruebel (approved):
Looks good to me

Issue #49, opened 2022-12-15
Fix support for external links on export


oruebel:
Exporting of HDF5 files with external links is not yet fully implemented/tested. ``tests/unit/test_io_zarr.py`` defines several test cases for this scenario that are not yet passing that would need to be addressed in order to complete support for external links on export:, e.g.,
* ``test_soft_link_dataset``
* ``test_external_link_group``
* ``test_external_link_dataset`` 
* ``test_external_link_link``
*  ``test_attr_reference``
*  ``test_append_data``
* ``test_append_external_link_data``
*  `` test_append_external_link_copy_data``
*  ``test_export_dset_refs``
*  ``test_export_cpd_dset_refs``, 

- ``hdmf.backends.hdf5.h5tools.HDF5IO`` uses the ``export_source`` argument on export. Need to check whether we may need to use it here as well to address this issue.

oruebel:
#120 added `export_source` so the parameter is present now. However, it is used there to enforce internal links, but the issue with external links remains. One place that #120 marks as a specific issue is the creation of paths in `Zarrio.__get_ref`

oruebel:
@mavaylon1 did you close this issue by accident or has this been fixed?

PR #120, opened 2023-09-16
Rework ZarrIO backend/Remove python3.7/Update HDMF and PyNWB min/Update workflows

mavaylon1:
## Motivation

This branch holds the re-work of the ZarrIO backend on how it handles object references. The changes are to follow the same workflow HDMF has when handling object references. 

With the changes to the backend, we have a temporary ignore on dtype conversion. It is has been raised in this ticket:  https://github.com/hdmf-dev/hdmf-zarr/pull/120

This PR removes python 3.7, brings hdmf and pynwb to the most up to date, as well updates to the workflows.

## How to test the behavior?
```
```

Run unit tests and gallery examples, specifically test_io_convert.py and plot_convert_nwb_hdf5.py 

## Checklist


src/hdmf_zarr/zarr_utils.py
```diff
@@ -0,0 +1,289 @@
+"""
+Utilities for the HDF5 I/O backend,
+e.g., for wrapping HDF5 datasets on read, wrapping arrays for configuring write, or
+writing the spec among others"""
```
oruebel:
```suggestion
Utilities for the Zarr I/O backend,
e.g., for wrapping Zarr arrays on read, wrapping arrays for configuring write, or
writing the spec among others
"""
```

src/hdmf_zarr/zarr_utils.py
```diff
@@ -0,0 +1,289 @@
...
+#from Zarrpy import filters as Zarrpy_filter
+from zarr import Array as ZarrArray
+
+from hdmf.array import Array
+from hdmf.query import HDMFDataset, ReferenceResolver, ContainerResolver, BuilderResolver
+from hdmf.utils import docval, getargs, popargs, get_docval
+
+
+class ZarrDataset(HDMFDataset):
+    @docval({'name': 'dataset', 'type': (np.ndarray, ZarrArray, Array), 'doc': 'the Zarr file lazily evaluate'},
```
oruebel: Add docstring for the class

src/hdmf_zarr/zarr_utils.py
```diff
@@ -0,0 +1,289 @@
...
+    """
+    A mixin for adding to Zarr reference-resolving types
+    the get_object method that returns Containers
+    """
+
+    def get_object(self, zarr_obj):
+        """
+        A class that maps an Zarr object to a Container
+        """
+        return self.io.get_container(zarr_obj)
```
oruebel: I think `BuilderResolverMixin` and `ContainerResolverMixin` are not implementing any Zarr-specific functionality. I think we may be able to just import these from HDMF. As the comment says, for import we may want to move them to `backend/utils.py` in HDMF instead

src/hdmf_zarr/zarr_utils.py
```diff
@@ -0,0 +1,289 @@
...
+    @property
+    def io(self):
+        return self.__io
+
+    @property
+    def shape(self):
+        return self.dataset.shape
+
+
+class DatasetOfReferences(ZarrDataset, ReferenceResolver, metaclass=ABCMeta):
```
oruebel, line 38: Aside from inheriting from `ZarrDataset` I think the functionality in this class is the same as in `DatasetOfReferences` in  HDF5. We may want to think about how we can avoid this duplication.

src/hdmf_zarr/zarr_utils.py
```diff
@@ -0,0 +1,289 @@
...
+    def resolve(self, manager):
+        return self[0:len(self)]
+
+    def __iter__(self):
+        for i in range(len(self)):
+            yield self[i]
+
+
+class AbstractZarrReferenceDataset(DatasetOfReferences):
+
```
oruebel: Add docstring for these classes.

For the classes for "Region" references, I would add a note that ZarrIO does not yet fully support region references and that the classes are, therefore, currently unused.

src/hdmf_zarr/backend.py
```diff
@@ -1132,91 +1149,45 @@ def __read_dataset(self, zarr_obj, name):
...
         if isinstance(dtype, list):
-            # compound data type
-            obj_refs = list()
-            reg_refs = list()
+            # Check compound dataset where one of the subsets contains references
+            has_reference = False
             for i, dts in enumerate(dtype):
-                if dts['dtype'] == DatasetBuilder.OBJECT_REF_TYPE:
-                    obj_refs.append(i)
+                if dts['dtype'] == 'object': # check items for object reference
```
```suggestion
                if dts['dtype'] in ['object', 'region']: # check items for object reference
```

src/hdmf_zarr/backend.py
```diff
@@ -1132,91 +1149,45 @@ def __read_dataset(self, zarr_obj, name):
...
                     has_reference = True
-                elif dts['dtype'] == DatasetBuilder.REGION_REF_TYPE:
-                    reg_refs.append(i)
-                    has_reference = True
-
+                    break
+                # TODO: Region reference not supported
+                # elif dts['dtype'] == 'region':
+                #     has_reference = True
+                #     break
```
```suggestion
```

src/hdmf_zarr/backend.py
```diff
@@ -1132,91 +1149,45 @@ def __read_dataset(self, zarr_obj, name):
...
+            retrieved_dtypes = [dtype_dict['dtype'] for dtype_dict in dtype]
+            if has_reference:
+                data = BuilderZarrTableDataset(zarr_obj, self, retrieved_dtypes)
```
```suggestion
            if has_reference:
                # TODO:  BuilderZarrTableDataset does not yet support region reference
                data = BuilderZarrTableDataset(zarr_obj, self, retrieved_dtypes)
```

src/hdmf_zarr/backend.py
oruebel: For compound datasets, `BuilderZarrTableDataset` will need to handle the region references, so I think we can hear prepare the code to already wrap the dataset correctly so that if/when we add region reference supprort we will only need to update `BuilderZarrTableDataset`

src/hdmf_zarr/backend.py
```diff
@@ -1132,91 +1149,45 @@ def __read_dataset(self, zarr_obj, name):
...
-                # TODO Should implement a lazy way to evaluate references for Zarr
-                data = deepcopy(data[:])
-                self.__parse_ref(kwargs['maxshape'], obj_refs, reg_refs, data)
-            except ValueError as e:
-                raise ValueError(str(e) + "  zarr-name=" + str(zarr_obj.name) + " name=" + str(name))
+            if dtype == 'object': # wrap with dataset ref
+                data = BuilderZarrReferenceDataset(data, self)
+            # TODO: Region reference not wrapped yet
+            # elif dtype == 'region':
+            #     reg_refs = True
```
```suggestion
            # TODO: Resolution of Region reference not yet supported by BuilderZarrRegionDataset
            elif dtype == 'region':
                data = BuilderZarrRegionDataset(data, self)
```

oruebel: You already have the wrapper classes for this setup, so I think we can do the wrapping here already. However, we should add a TODO item in the utils wherever region references need to be resolved and maybe raise a NotImplementedError if appropriate.

src/hdmf_zarr/backend.py
```diff
@@ -563,11 +579,13 @@ def __get_ref(self, ref_object):
...
-                  else self.source)
-        # Make the source relative to the current file
-        source = os.path.relpath(os.path.abspath(source), start=self.abspath)
+        # source = (builder.source
+        #           if (builder.source is not None and os.path.isdir(builder.source))
+        #           else self.source)
+        #
+        # # Make the source relative to the current file
+        # source = os.path.relpath(os.path.abspath(source), start=self.abspath)
+        source = '.'
```
mavaylon1: @oruebel What Ryan and I found in HDMF was that the source is always the file itself and not an external source, in this case the builder.source is the file which it was written. The equivalent was what I did, which was to say `source = '.'`
Though HDF5 supports external references, we have references that are only internal.

This plus an update to resolve_ref was our fix until we add the "queue" mechanism of writing/exporting references in the future.

src/hdmf_zarr/backend.py
```diff
@@ -531,7 +541,13 @@ def __resolve_ref(self, zarr_ref):
             try:
                 target_zarr_obj = target_zarr_obj[object_path]
             except Exception:
-                raise ValueError("Found bad link to object %s in file %s" % (object_path, source_file))
+                try:
+                    import pathlib
```
mavaylon1: @oruebel This is the update to `resolve_ref` I mentioned below. This checks that the relative_path works to retrieve the target. This would change/be updated as we bring zarr up to speed with the "queue" HDMF uses so that we don't need this check.

src/hdmf_zarr/zarr_utils.py
mavaylon1: Agreed. I think for now this is fine because I don't want to slow down a release on this next week on waiting for a HDMF ticket to be merged to support this change. I will make an issue ticket on HDMF still with the changes needed and a ticket on Zarr to make sure we update this once that has been merged. Timeline for that should be before Boston still as it shouldn't take long at all, that being said I don't want it slowing down this release. This is the same for the other comments you mentioned about refactoring to the backend in HDMF. @oruebel

src/hdmf_zarr/backend.py
oruebel: Can you clarify what you mean by "queue mechanism for export?

mavaylon1: @oruebel Currently in hdmf, references are resolved once everything else have been written. They are stored in a queue. 
```
@self.__queue_ref
                    def _filler():
                        self.logger.debug("Resolving object references and setting attribute on dataset '%s' "
                                          "containing attributes: %s"
                                          % (name, list(attributes.keys())))
                        refs = list()
                        for item in data:
                            refs.append(self.__get_ref(item))
                        dset = parent[name]
                        dset[()] = refs
                        self.set_attributes(dset, attributes)```
This guarantees that they will be resolved. This filler method will "fill in" on write with dset[()] = refs.

From my understanding, there is no set order to which objects are written, leaving the possibility that some references won't be properly resolved. Having them pushed to a queue to the end guarantees this resolution.

Comment (oruebel) [reply src/hdmf_zarr/backend.py]: Makes sense. Thanks for the clarification. I think this was probably added after ZarrIO was branched from HDF5IO, probably for export and that this has not made it in ZarrIO yet.

**Comment by mavaylon1** [comment]
Fix #121 
Fix #113
Fix #106 
Fix #79

**Comment by mavaylon1** [comment]
@rly I did the updates for using export_source for a flag; however, we would need to flush it out more with the export updates in the future. (specifically in links). The scope of these this PR is getting to large.

File: src/hdmf_zarr/backend.py [hunk 12]
```diff
@@ -566,8 +589,12 @@ def __get_ref(self, ref_object):
         source = (builder.source
                   if (builder.source is not None and os.path.isdir(builder.source))
                   else self.source)
+
         # Make the source relative to the current file
         source = os.path.relpath(os.path.abspath(source), start=self.abspath)
+
+        if export_source is not None:
+            source = '.'
```
oruebel: For future reference, I would add a TODO item to clarify what needs to be done here in the future. I think all we really need is a way to figure out whether the target for the reference will be copied into the new file on export or whether it will remain only in the source we are exporting from. Otherwise, resolution of external links is I believe already being handled. 

```suggestion
        # TODO: This check assumes that all links are internal links on export. Need to deal with external links on export.
        if export_source is not None:
            source = '.'
```

src/hdmf_zarr/backend.py
```diff
@@ -344,7 +359,8 @@ def write_group(self, **kwargs):
                 self.write_dataset(parent=group,
                                    builder=sub_builder,
                                    link_data=link_data,
-                                   exhaust_dci=exhaust_dci)
+                                   exhaust_dci=exhaust_dci,
+                                   export_source=export_source)
```
oruebel, line 362: @mavaylon1 references are also stored in attributes. We will need to pass `export_source` also to `write_attributes` so that it can pass the `export_source` also in its calls to `__get_ref`

oruebel: You may want to search for `__get_ref` to make sure that all places where `__get_ref` is used on write are passed the `export_source` parameter.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## 0.3.1 (Upcoming)
 
+### Enhancements
+* Reworked ZarrIO to resolve object references in the same manner as HDMF, updated HDMF and PyNWB version to the most recent release, removed python 3.7 as the minimum. @mavaylon1 [#120](https://github.com/hdmf-dev/hdmf-zarr/pull/120)
```
```suggestion
### Enhancements
* Enhanced ZarrIO to resolve object references lazily on read similar to HDMF's `HDF5IO` backend @mavaylon1 [#120](https://github.com/hdmf-dev/hdmf-zarr/pull/120)

### Dependencies
* Updated HDMF and PyNWB version to the most recent release @mavaylon1 [#120](https://github.com/hdmf-dev/hdmf-zarr/pull/120)
* Updated minimum Python version from 3.7 to 3.8 @mavaylon1 [#120](https://github.com/hdmf-dev/hdmf-zarr/pull/120)
```

src/hdmf_zarr/backend.py
mavaylon1: Ah missed those. Good catch

src/hdmf_zarr/backend.py
```diff
@@ -531,11 +547,18 @@ def __resolve_ref(self, zarr_ref):
...
             except Exception:
-                raise ValueError("Found bad link to object %s in file %s" % (object_path, source_file))
+                # TODO: Find why the reference includes the root object.
+                # This is a fix to allow the target object to be found
+                try:
+                    object_path = pathlib.Path(object_path)
+                    rel_obj_path = object_path.relative_to(*object_path.parts[:2])
+                    target_zarr_obj = target_zarr_obj[rel_obj_path]
+                except Exception:
+                    raise ValueError("Found bad link to object %s in file %s" % (object_path, source_file))
```
oruebel: With the fix in `__get_ref` using `export_source`, this workaround on read should no longer be necessary

src/hdmf_zarr/backend.py
```diff
@@ -783,29 +818,31 @@ def write_dataset(self, **kwargs):  # noqa: C901
...
-            if isinstance(data, RegionBuilder):
-                shape = (1,)
-                type_str = 'region'
-                refs = self.__get_ref(data.builder, data.region)
-            elif isinstance(data, ReferenceBuilder):
+            # TODO: Region References are not yet supported
+            # if isinstance(data, RegionBuilder):
+            #     shape = (1,)
+            #     type_str = 'region'
+            #     refs = self.__get_ref(data.builder, data.region)
```
```suggestion
```

Comment is the same as the code below it

```suggestion
```

The first part of the `if` statement appears to refer to `object` references not `region` references

src/hdmf_zarr/backend.py
mavaylon1: @oruebel we do need this still. The source is correct, but there actual path includes the source. This is a fix to remove the source. This would be a bandage until the next update on the backend to flush out export, but also the paths for references.

oruebel: That seems like an issue that should be addressed in `__get_ref`. We are correcting the `source` in `__get_ref`. We'll probably need to do something similar for `path`. We shouldn't create files with bad paths.

oruebel:
We should also update the note here https://github.com/hdmf-dev/hdmf-zarr/blob/dev/docs/source/storage.rst#region-references to mention that the new resolver classes will also need to be updated to implement region references

mavaylon1:
> We should also update the note here https://github.com/hdmf-dev/hdmf-zarr/blob/dev/docs/source/storage.rst#region-references to mention that the new resolver classes will also need to be updated to implement region references

Done.

.github/workflows/run_all_tests.yml
```diff
@@ -162,7 +158,7 @@ jobs:
         run: |
           conda config --set always_yes yes --set changeps1 no
           conda info
-          # the conda dependency resolution for tox under python 3.7 can install the wrong importlib_metadata
+          # the conda dependency resolution for tox under python 3.8 can install the wrong importlib_metadata
           conda install -c conda-forge tox "importlib_metadata>4"
```
rly:
```suggestion
          conda install -c conda-forge tox
```
The importlib_metadata install from conda-forge was necessary under python 3.7 but is no longer needed if we drop testing on python 3.7

src/hdmf_zarr/backend.py
```diff
@@ -566,8 +584,15 @@ def __get_ref(self, ref_object):
...
                   else self.source)
+
         # Make the source relative to the current file
         source = os.path.relpath(os.path.abspath(source), start=self.abspath)
+        # TODO: This check assumes that all links are internal links on export.
+        # Need to deal with external links on export.
+        if export_source is not None:
+            # Make sure the source of the reference is now towards the new file
+            # and not the original source when exporting.
+            source = '.'
```
```suggestion
        # TODO: This check assumes that all links are internal links on export.
        # Need to deal with external links on export.
        if export_source is not None:
            # Make sure the source of the reference is now towards the new file
            # and not the original source when exporting.
            source = '.'
        else:
            source = os.path.relpath(os.path.abspath(source), start=self.abspath)
```

tests/unit/test_io_convert.py
```diff
@@ -163,6 +169,22 @@ def test_export_roundtrip(self):
                     container=container,
                     write_path=write_path,
                     export_path=export_path)
+                if self.REFERENCES:
+                    if self.TARGET_FORMAT == "H5":
+                        num_bazs = 10
+                        for i in range(num_bazs):
+                            baz_name = 'baz%d' % i
+                            self.assertEqual(exported_container.baz_data.data.__class__.__name__,
+                                             'ContainerH5ReferenceDataset')
```
rly: Can we evaluate the classes are equal instead of the class name? That's more typical. This formulation makes it look like there is a reason we cannot import these classes. e.g.,
`self.assertIsInstance(exported_container.baz_data.data, ContainerH5ReferenceDataset)`

rly:
This looks good to me. I'll let @oruebel do the final approval.

rly:
Thanks for your hard work on this!

.readthedocs.yaml
```diff
@@ -21,6 +26,7 @@ python:
   install:
     - requirements: requirements-doc.txt
     - requirements: requirements.txt
+    - path: . # path to the package relative to the root
```
oruebel, line 29: Just curious. Why is specifying `path` necessary?

mavaylon1: What ended up happening was the build was failing. I looked at hdmf to see how it was set up there. Looking at the error, it was not properly installing the requirements when building the docs. I looked at the documentation for how to setup this yaml and it says the "path" would be the path to the requirements file. I confirmed that in hdmf it was also present. These changes allowed for the docs to properly find the requirements and be built.

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## 0.3.1 (Upcoming)
```
```suggestion
## 0.4.0 (Upcoming)
```

This seems more than just a bug-fix release. So I would recommend bumping the version to 0.4

docs/gallery/plot_convert_nwb_hdf5.py
```diff
@@ -107,9 +105,10 @@
 #
 # Using the same approach as above, we can now convert our Zarr file back to HDF5.
 
-with NWBZarrIO(zarr_filename, mode='r') as read_io:  # Create Zarr IO object for read
-    with NWBHDF5IO(hdf_filename, 'w') as export_io:  # Create HDF5 IO object for write
-        export_io.export(src_io=read_io, write_args=dict(link_data=False))  # Export from Zarr to HDF5
+with suppress(Exception):  # TODO: This is a temporary ignore on the convert_dtype exception.
```
oruebel, line 108: @mavaylon1 can you please create an issue for this so we don't forget to check on the dtype conversion issue (and include links to the lines where the issue is being suppressed)

docs/source/storage.rst
```diff
@@ -282,6 +282,8 @@ region references link to subsets of another Dataset. To identify region referen
 to the ``source`` and ``path``, the  py:class:`~hdmf_zarr.utils.ZarrReference` object will also need to
 store the definition of the ``region`` that is being referenced, e.g., a slice or list on point indices.
 
+In the future, the reference resolver classes will need to be updated to implement region references
```
oruebel: @mavaylon1  can you please move this line to the list of TODO items in the warning message right below it and add mention where the reference resolver classes are located.

src/hdmf_zarr/backend.py
```diff
@@ -566,8 +584,16 @@ def __get_ref(self, ref_object):
         source = (builder.source
                   if (builder.source is not None and os.path.isdir(builder.source))
                   else self.source)
+
         # Make the source relative to the current file
-        source = os.path.relpath(os.path.abspath(source), start=self.abspath)
+        # TODO: This check assumes that all links are internal links on export.
```
oruebel, line 589: Just FYI, looking through issues, we already have an open issue on export with external links here https://github.com/hdmf-dev/hdmf-zarr/issues/49  I.e., any specific issues we identify here related to export with external links, we should document there. I added a note, about this particular change in `__get_ref` to the issue, but if there are other items that you identified in this PR, then please add notes to that issue.

src/hdmf_zarr/backend.py
```diff
@@ -783,29 +817,26 @@ def write_dataset(self, **kwargs):  # noqa: C901
...
                 # write a compound datatype
                 dset = self.__list_fill__(parent, name, data, options)
         # Write a dataset of references
         elif self.__is_ref(options['dtype']):
-            if isinstance(data, RegionBuilder):
-                shape = (1,)
-                type_str = 'region'
-                refs = self.__get_ref(data.builder, data.region)
-            elif isinstance(data, ReferenceBuilder):
+            if isinstance(data, ReferenceBuilder):
```
oruebel, line 832:
```suggestion
            # TODO Region references are not yet support, but here how the code should look
            #  if isinstance(data, RegionBuilder):
            #      shape = (1,)
            #      type_str = 'region'
            #      refs = self.__get_ref(data.builder, data.region)
            if isinstance(data, ReferenceBuilder):
```

I know I originally suggested removing this comment. But I would keep it, because it will make implementing region references easier in the future.

src/hdmf_zarr/backend.py
```diff
@@ -1229,3 +1226,16 @@ def __read_attrs(self, zarr_obj):
                 else:
                     ret[k] = v
         return ret
+
+    @docval({'name': 'zarr_obj', 'type': (Array, Group),
+             'doc': 'the Zarr object to the corresponding Container/Data object for'})
+    def get_container(self, **kwargs):
```
oruebel: Very minor. I would move the function to make sure `get_container` and `get_builder` are in the same place in the code, since the two are related.

src/hdmf_zarr/zarr_utils.py
oruebel: Agreed, creating issues for this is fine.

tests/unit/utils.py
```diff
@@ -306,8 +306,9 @@ class BazCpdData(Data):
 
 
 class BazBucket(Container):
+    ROOT_NAME = 'root'
```
oruebel: For future proofing, I would just use the ROOT_NAME from ZarrIO (which I believe is already being imported). Similarly, in the docval, the default for 'name' can then be ROOT_NAME as imported from ZarrIO instead of redefining 'root' as a string.

mavaylon1: I don't see that we import ZarrIO to this file. This is also set in FooFile (like in HDMF), I will define it outside the class and use ROOT_NAME

docs/gallery/plot_convert_nwb_hdf5.py
mavaylon1: Done

tests/unit/test_io_convert.py
```diff
@@ -163,6 +171,20 @@ def test_export_roundtrip(self):
...
+                            baz_name = 'baz%d' % i
+                            self.assertIsInstance(exported_container.baz_data.data, ContainerH5ReferenceDataset)
+                            self.assertIs(exported_container.baz_data.data[i], exported_container.bazs[baz_name])
+                    elif self.TARGET_FORMAT == "ZARR":
+                        num_bazs = 10
+                        for i in range(num_bazs):
+                            baz_name = 'baz%d' % i
+                            self.assertIsInstance(exported_container.baz_data.data, ContainerZarrReferenceDataset)
+                            self.assertIs(exported_container.baz_data.data[i], exported_container.bazs[baz_name])
+
```
oruebel, line 187: @mavaylon1 according to Codecov, this code seems to not be hit during the tests. It seems strange that the references appear as not tested with "ZARR" as a the target format. 

Screen Shot 2023-09-29 at 3 43 10 PM

Can you please double-check why this is not being used? The CodeCov for this PR is here https://app.codecov.io/gh/hdmf-dev/hdmf-zarr/pull/120

tests/unit/test_io_convert.py
```diff
@@ -196,6 +217,7 @@ class MixinTestHDF5ToZarr():
                     DirectoryStore('test_export_DirectoryStore.zarr'),
                     TempStore(),
                     NestedDirectoryStore('test_export_NestedDirectoryStore.zarr')]
+    TARGET_FORMAT = "Zarr"
```
oruebel:
```suggestion
    TARGET_FORMAT = "ZARR"
```

@mavaylon1  based on the code above, I believe this should be "ZARR" (all uppercase). I think this is likely why the check for references does not appear to be run in the tests.

tests/unit/utils.py
```diff
@@ -11,6 +11,7 @@
 from hdmf.utils import (docval, getargs, get_docval)
 
 CORE_NAMESPACE = 'test_core'
+ROOT_NAME = 'root'
```
oruebel: Is there a reason why we would not just import it?

```suggestion
from hdmf_zarr.backend import ROOT_NAME
```

tests/unit/test_io_convert.py
mavaylon1: Done

tests/unit/utils.py
mavaylon1: I don't think so. I think the manual definition within utils.py is a legacy carry over from HDMF, where it is set with all the Foo and Baz containers and not imported.

tests/unit/test_io_convert.py
```diff
@@ -259,6 +282,7 @@ class MixinTestZarrToZarr():
                     DirectoryStore('test_export_DirectoryStore_Export.zarr'),
                     TempStore(dir=os.path.dirname(__file__)),   # set dir to avoid switching drives on Windows
                     NestedDirectoryStore('test_export_NestedDirectoryStore_Export.zarr')]
+    TARGET_FORMAT = "Zarr"
```
oruebel: Same as above. I believe this needs to be "ZARR" (all caps) for the checks for references to be run.

```suggestion
    TARGET_FORMAT = "ZARR"
```

oruebel (approved):
Looks good to me. Thanks for your patients and hard work on this PR.

PR #129, opened 2023-10-01
Workflows

mavaylon1:
## Motivation

Add workflows that exist in HDMF: ruff (to replace flake8) and codespell.

## How to test the behavior?
```
```
Run the workflows

## Checklist

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,19 @@
+---
+name: Codespell
+
+on:
+  push:
+    branches: [dev]
+  pull_request:
+    branches: [dev]
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Codespell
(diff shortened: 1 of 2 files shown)


mavaylon1:
@rly @oruebel I'm adding these workflows and I suppose a larger question is coming up: "Should this PR also include the shift from setup.py to pyproject.toml/remove versioneer to how hdmf handles versions with hatch". I ask this because flake8 is setup through the .cfg and HDMF replaced flake8 with ruff (which is handled in the pyproject.toml). I'm thinking that this PR can be:
1. Bring the workflows up to HDMF
2. switch to pyproject.toml
3. Switch version control to hatch
The only thing that this would wait for is the fix to the release bug I showed @rly on Friday that creates a wheel with the wrong version. Once that fix is in HDMF, that can also be included here. I think this can be done this week depending on the release bug. I also created a ticket for the bug: https://github.com/hdmf-dev/hdmf/issues/956

mavaylon1:
This PR is not ready for review, nor have I finished supporting these new workflows.

rly:
Yes, let's do a full update so that the testing and maintenance infrastructure is the same as in hdmf.

Issue #126, opened 2023-10-01
[Feature]: Run tests again dev branch of HDMF


oruebel:
### What would you like to see added to HDMF-ZARR?

To detect errors due to changes in the dev branch of HDMF it will be useful to run tests against the lastest HDMF on the dev branch. These tests should not be required for merge.

### Is your feature request related to a problem?

The test suite currently fails when running against HMDF on the dev branch

### What solution would you like?

...

### Do you have any interest in helping implement the feature?

No.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/hdmf-dev/hdmf-zarr/blob/dev/.github/CODE_OF_CONDUCT.md)
- [X] Have you checked the [Contributing](https://github.com/hdmf-dev/hdmf-zarr/blob/dev/docs/CONTRIBUTING.rst) document?
- [X] Have you ensured this change was not already [requested](https://github.com/hdmf-dev/hdmf-zarr/issues)?

mavaylon1:
You want to have a workflow here that runs HDMF-Zarr but using the dev branch of HDMF rather than the released HDMF correct? @oruebel

oruebel:
Correct

mavaylon1:
> Correct

Will do

PR #136, opened 2023-11-01
Create HDMF_dev.yaml

mavaylon1:
## Motivation

This introduces a new workflow to run the tests on HDMF-Zarr using the dev branch of HDMF. 

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/.github/workflows/HDMF_dev.yaml
+++ b/.github/workflows/HDMF_dev.yaml
@@ -22,14 +22,15 @@ jobs:
       - name: Install HDMF_Zarr Requirements
         run: |
           python -m pip install -r requirements-dev.txt -r requirements.txt
+          python -m pip install -e .
           
-      # - name: Clone HDMF Dev Branch
-      #   run: |
-      #     git clone https://github.com/hdmf-dev/hdmf.git --recurse-submodules
-      #     cd hdmf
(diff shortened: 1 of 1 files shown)


mavaylon1:
Fix #126

mavaylon1:
@oruebel As I was making this, I started to think about how much did we want to test. Within tox.ini, it installs the latest release of HDMF. It will run pytest according to different environment versions. 

Now ideally, this workflow could do something like that, but we can't use tox.ini because we would have to edit the install requirements to use the dev branch. What I have now is a proof of concept that it works for only python 3.11. Do we want this workflow to only do that or should we get creative and test for all python versions?

rly:
@mavaylon1 I think testing on python 3.11 (or whatever is the latest version supported by both packages) is sufficient. We could test this workflow on all python versions, on all OSs, with minimum, pinned, upgraded, and pre-release requirements, but I think that is overkill and unlikely to catch issues.

rly:
Unlikely to catch any issues that testing on each package independently would not catch.

mavaylon1:
> Unlikely to catch any issues that testing on each package independently would not catch.

Agreed.

mavaylon1:
Since this is a workflow change, I am not updating the Changelog

rly (approved):
LGTM

Issue #139, opened 2023-11-10
[Feature]: Always consolidate store structure


CodyCBakerPhD:
### What would you like to see added to HDMF-ZARR?

From discussion on https://github.com/dandi/dandi-archive/issues/1745#issuecomment-1806440234

Zarr command: https://zarr.readthedocs.io/en/stable/api/convenience.html#zarr.convenience.consolidate_metadata

Always enable consolidation of metadata on NWB file creation to make it easier to traverse store contents and on S3, reduce the number of head-like requests that would be needed to read store structure

### What solution would you like?

Wish the NWBZarrIO or higher level ZarrIO would do this automatically on file creation, but unsure where in the stack to inject that

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/hdmf-dev/hdmf-zarr/blob/dev/.github/CODE_OF_CONDUCT.md)
- [X] Have you checked the [Contributing](https://github.com/hdmf-dev/hdmf-zarr/blob/dev/docs/CONTRIBUTING.rst) document?
- [X] Have you ensured this change was not already [requested](https://github.com/hdmf-dev/hdmf-zarr/issues)?

oruebel:
Interesting, I wasn't aware of this feature. The call to create the consolidated metadata I think would need to happen at the end of 'ZarrIO.write_builder' and then 'ZarrIO.open' wound need to check if the consolidated metadata exist to use 'open_consoludated' instead. We'd probably want to also add a 'ZarrIO.consolidate_metadata(..)' function to explicitly update/generate the metadata file and add a flag on write to allow to enable/disable on write.

oruebel:
@mavaylon1 do you think you can take a stab at this?

mavaylon1:
> @mavaylon1 do you think you can take a stab at this?

I shall.

PR #142, opened 2023-11-21
Consolidate Metadata

mavaylon1:
## Motivation

Addressing #139. By default, the metadata will be consolidated into one file, `.zmetadata`. This will check the existence of this file and will still open the file even if the metadata is not consolidated. 

- [x] Boolean for consolidating metadata
- [x] Build tests for when metadata is and is not consolidated
- [x] General zarr file builder for tests
- [x] Update docs with notes on metadata updates.

## How to test the behavior?
```
```

## Checklist


mavaylon1:
Fix #139

src/hdmf_zarr/backend.py
```diff
@@ -1202,7 +1261,11 @@ def read_builder(self):
         return f_builder
 
     def __set_built(self, zarr_obj, builder):
-        fpath = zarr_obj.store.path
+        # fpath = zarr_obj.store.path
+        if isinstance(zarr_obj.store, zarr.storage.ConsolidatedMetadataStore):
+            fpath = zarr_obj.store.store.path
+        else:
+            fpath = zarr_obj.store.path
```
oruebel: Since you use this logic of having to distinguish by the the storage type to determine the path, I think it would be useful to make a small private helper function for this, e.g,. ``__get_store_path``

mavaylon1: Agreed. This was a checkpoint push since I got everything to pass. I need to optimize and finish some tests.

src/hdmf_zarr/backend.py
```diff
@@ -157,10 +157,19 @@ def object_codec_class(self):
     def open(self):
         """Open the Zarr file"""
         if self.__file is None:
-            self.__file = zarr.open(store=self.path,
-                                    mode=self.__mode,
-                                    synchronizer=self.__synchronizer,
-                                    storage_options=self.__storage_options)
+            # Within zarr, open_consolidated only allows the mode to be 'r' or 'r+'.
+            # As a result, when in 'w' mode, the file will not use consolidated metadata.
+            if self.__mode == 'w':
```
oruebel: What about modes `'a'` and  `'w-'`? Since open_cosolidated only allows for 'r' and 'r+' it may be easier to check for `if self.__mode not in ['r', 'r+']`

src/hdmf_zarr/backend.py
```diff
@@ -426,6 +434,61 @@ def write_builder(self, **kwargs):
...
+        When a file is written, the metadata within the file is consolidated automatically.
+        If there are any metadata changes, the user needs to consolidate the metadata again
+        with this method in order for the metadata to be read correctly.
+
+        Consolidate all metadata for groups and arrays within the given store into a
+        single resource and put it under .zmetadata.
+        """
+        zarr.consolidate_metadata(store, metadata_key='.zmetadata')
+
+    def __get_store_path(self, zarr_object):
```
oruebel: Instead changing `get_zarr_parent_path` and  `get_zarr_paths`to non-static method, I think we can simply make this a static method. It looks like the `self` argument is not needed here, so making this function static should be fine. 

```suggestion
    @staticmethod
    def __get_store_path(zarr_object):
```

mavaylon1: Yep already did that locally.

mavaylon1:
Comments for review
1. There are a few instances in backend.py that remain `zarr.open()` rather than `__open_consolidated`.  

- In Line 81, I left this as a simple test of whether it can be read. 
- In Line 199, the goal is to load the cached namespace from a file. The parameters in this `zarr.open()` don't include any of the optional arguments for `__open_consolidated`, and the whole point of having a consolidated metadata is for quick look-ups for groups, datasets, etc (nothing to do with the namespace). As a result, I think it might be "safer" (meaning less likely for a bug to affect this area) if we kept it as is. That being said it is an easy switch if the majority disagrees with this view.

2. Why are there only two tests? With the `zarr.consolidate_metadata` being called at the end of write_builder, it will generate test examples (including examples with references) that have `zmetadata`, meaning it's covered by existing tests. (As seen in coverage).

3. Why is there no `consolidate_metadata` method anymore? This method was here briefly in this PR, but I removed it since it is literally calling the same method within zarr. I think wrapping it with the same name would just create confusion if something goes wrong, meaning the issue is not us but zarr. I don't want us to have issue tickets saying "it's broken" just because they are not using the zarr method correctly.

tests/unit/test_zarrio.py
```diff
@@ -122,3 +129,24 @@ class TestExportZarrToZarrNestedDirectoryStore(BaseTestExportZarrToZarr):
...
+
+
+#########################################
+#  Consolidate Metadata tests
+#########################################
+class TestConsolidateMetadata(TestCase):
+    """
+    Tests for consolidated metadata and corresponding helper methods.
+    """
+    def test_get_store_path_shallow(self):
```
oruebel, line 141: Should there be a ``def tearDown()`` method here to delete the `'tests/unit/example.zarr'` and `'tests/unit/test_consolidate.zarr'` files? And related to that, are those two files required to be stored in the repo or should they be removed from the PR? It looks like the files are automatically generated each time the test-suite is run, so it seems they are probably not needed.

mavaylon1: I actually wanted to keep them, similar to how we keep the termset example for hdmf. It seemed easier to have a set of test files I could always provide a path to for future tests that can be in different test scripts.

oruebel:
Thanks for the additional comments.

> 1. There are a few instances in backend.py that remain `zarr.open()` rather than `__open_consolidated`.
> 
> 
>     * In Line 81, I left this as a simple test of whether it can be read.
> 
>     * In Line 199, the goal is to load the cached namespace from a file.

I agree, it makes sense to leave `zarr.open` in those two instances. 

> 2\. Why are there only two tests? With the `zarr.consolidate_metadata` being called at the end of write_builder, it will generate test examples (including examples with references) that have `zmetadata`, meaning it's covered by existing tests. (As seen in coverage).

Makes sense. This raises the question whether we need some flag in the tests to run them with and without consolidation of metadata. I.e., I assume now those tests always run with consolidated metadata but no longer test for the case where metadata is not consolidated. 


> 3\. Why is there no `consolidate_metadata` method anymore? This method was here briefly in this PR, but I removed it since it is literally calling the same method within zarr. I think wrapping it with the same name would just create confusion if something goes wrong, meaning the issue is not us but zarr. I don't want us to have issue tickets saying "it's broken" just because they are not using the zarr method correctly.

When would the user directly need to call `consolidate_metadata`? If there is a case where a user will need to call this method in zarr, then I think it would be useful to mention that somewhere in the docs.

mavaylon1:
> Thanks for the additional comments.
> 
> > 1. There are a few instances in backend.py that remain `zarr.open()` rather than `__open_consolidated`.
> >    
> >    * In Line 81, I left this as a simple test of whether it can be read.
> >    * In Line 199, the goal is to load the cached namespace from a file.
> 
> I agree, it makes sense to leave `zarr.open` in those two instances.
> 
> > 2. Why are there only two tests? With the `zarr.consolidate_metadata` being called at the end of write_builder, it will generate test examples (including examples with references) that have `zmetadata`, meaning it's covered by existing tests. (As seen in coverage).
> 
> Makes sense. This raises the question whether we need some flag in the tests to run them with and without consolidation of metadata. I.e., I assume now those tests always run with consolidated metadata but no longer test for the case where metadata is not consolidated.
> 
> > 3. Why is there no `consolidate_metadata` method anymore? This method was here briefly in this PR, but I removed it since it is literally calling the same method within zarr. I think wrapping it with the same name would just create confusion if something goes wrong, meaning the issue is not us but zarr. I don't want us to have issue tickets saying "it's broken" just because they are not using the zarr method correctly.
> 
> When would the user directly need to call `consolidate_metadata`? If there is a case where a user will need to call this method in zarr, then I think it would be useful to mention that somewhere in the docs.


2. Well with this PR, it will always be consolidated. I think it raises a question for how many zarr files are in the wild already (assuming with the Allen Institute) that are from hdmf-zarr prior to this PR. Even so, I can use `example.zarr` for a test for this case.
4. The user would need to call it again if they edit the file after it has been written. I can make a quick tutorial for sure.

tests/unit/test_zarrio.py
oruebel: The key difference is that TermSet are not generated or modified by the tests but only used. Here, however, the files are a main component of the test, which means: 

- if a test modifies the files then that leads to many updates and cross-talk between tests that are hard to debug.
- if we need to update the files in the future, there is no record how these files where generated

As an alternative, I think it may be more robust to add a function to  `https://github.com/hdmf-dev/hdmf-zarr/blob/dev/tests/unit/utils.py` to create (and delete) the test files so that it can be easily generated and cleaned-up by the tests that need a basic test file. In this way it is also explicit how these files are being generated.

mavaylon1: Sounds good.

src/hdmf_zarr/backend.py
```diff
@@ -426,6 +434,49 @@ def write_builder(self, **kwargs):
         self.logger.debug("Done writing %s '%s' to path '%s'" %
                           (f_builder.__class__.__qualname__, f_builder.name, self.source))
 
+        # Consolidate metadata for the entire file after everything has been written
+        zarr.consolidate_metadata(store=self.path)
```
oruebel: I'm wondering whether we should have a boolean option (that is True by default) on `write` (and `write_builder`) that a user can set when calling `write` to enable/disable consolidation of metadata. I think in most cases we'll want to consolidated metadata but depending on how long this takes, it may also be something that a user may want to be able to disable. What do you think?

oruebel:
Can you also add a brief section on consolidation of metadata to https://github.com/hdmf-dev/hdmf-zarr/blob/dev/docs/source/storage.rst so that we have a record in the docs on where this is stored and why.

oruebel:
> 2\. Well with this PR, it will always be consolidated.

I don't think we want to make the hard assumption that metadata will always be consolidated.

src/hdmf_zarr/backend.py
mavaylon1: I mean in practice, having this file will allow for a more efficient metadata retrieval on read. I mean I think there is no harm in having both, but I don't think it should cost a significant amount of time to generate this consolidation file. I think it is a good thing for the user to have, but the option is fine. I will create a bool, but the default will be True.

oruebel: Agreed.

docs/source/storage.rst
```diff
@@ -394,3 +394,10 @@ data type. The specification of the namespace is stored in
 ``/specifications/<namespace-name>/<version>/<source-filename>``. Here ``<source-filename>`` refers to the main name
 of the source-file without file extension (e.g., the core namespace defines ``nwb.ephys.yaml`` as source which would
 be stored in ``/specifications/core/2.0.1/nwb.ecephys``).
+
+Consolidating Metadata
+======================
+
+Zarr allows users to consolidate all metadata for groups and arrays within the given store. By default, every file
+will consolidate all metadata within into a single `.zmetadata` file, stored in the root group. This reduces the number of read
+operations when retrieving certain metadata in read mode.
```
oruebel, line 403:
```suggestion
operations when retrieving certain metadata in read mode.

.. note::

    When updating a file, the consolidated metadata will also need to be updated via 
    `zarr.consolidate_metadata(path)` to ensure the consolidated metadata is consistent 
    with the file.

```

src/hdmf_zarr/backend.py
```diff
@@ -238,6 +246,14 @@ def load_namespaces(cls, namespace_catalog, path, namespaces=None):
             ),
             "default": None,
         },
+        {
+            "name": "consolidate_metadata",
+            "type": bool,
+            "doc": (
+                "Boolean to consolidate metadata into a single a .zmetadata file within root group."
```
```suggestion
                "Boolean to consolidate metadata into a single .zmetadata file in the root group to accelerate read."
```

src/hdmf_zarr/backend.py
```diff
@@ -238,6 +246,14 @@ def load_namespaces(cls, namespace_catalog, path, namespaces=None):
             ),
             "default": None,
         },
+        {
+            "name": "consolidate_metadata",
+            "type": bool,
+            "doc": (
+                "Boolean to consolidate metadata into a single .zmetadata file in the root group to accelerate read."
```
```suggestion
                "Consolidate metadata into a single .zmetadata file in the root group to accelerate read."
```

src/hdmf_zarr/backend.py
```diff
@@ -398,11 +414,19 @@ def get_builder_disk_path(self, **kwargs):
             'doc': 'The source of the builders when exporting',
             'default': None,
         },
+        {
+            "name": "consolidate_metadata",
+            "type": bool,
+            "doc": (
+                "Boolean to consolidate metadata into a single a .zmetadata file within root group."
```
```suggestion
                "Consolidate metadata into a single .zmetadata file in the root group to accelerate read."
```

src/hdmf_zarr/backend.py
```diff
@@ -426,6 +450,50 @@ def write_builder(self, **kwargs):
...
+        This method will check to see if the metadata has been consolidated.
+        If so, use open_consolidated.
+        """
+        # self.path can be both a string or a one of the `SUPPORTED_ZARR_STORES`.
+        if isinstance(self.path, str):
+            path = self.path
+        else:
+            path = self.path.path
+
+        if os.path.isfile(path+'/.zmetadata'):
```
oruebel, line 486:
```suggestion
        if os.path.isfile(path+'/.zmetadata') and mode not in ['r', 'r+']:
```

Do we also need to check for the supported mode here?

oruebel (commented):
This looks overall good to me. One quick question, are we still running any tests now with `consolidate_metadata=False` or all tests now run with consolidated metadata only?

src/hdmf_zarr/backend.py
mavaylon1: No. There's only two places it's used. One has the check (in open) and the other is in read mode.

mavaylon1:
> This looks overall good to me. One quick question, are we still running any tests now with `consolidate_metadata=False` or all tests now run with consolidated metadata only?

There's one test to check that it runs with False, in relation to consolidated metadata. However, the rest is using the default True. I don't think it's a big enough change to test for the entire test suite for both but thoughts?

oruebel:
> I don't think it's a big enough change to test for the entire test suite for both but thoughts?

I think that would be nice, if only because it would help safeguard against potential issues as we make changes in the future. That being said, configuring the test suite to run in both modes is probably a non-trivial undertaking and is not worth holding up this PR for. Please just create an issue for future reference and we can decide separately on how to proceed on that question.

oruebel (approved):

Issue #145, opened 2023-12-08
[Bug]: Extension namespaces not exported from HDF5 NWB


sneakers-the-rat:
### What happened?

Testing an export of [a dataset](https://api.dandiarchive.org/api/dandisets/000021/versions/draft/assets/4513d0c9-1e2b-4c8a-aa22-ae81822537c9/) that has a custom namespace/schema, and found that it isn't exported and so a subsequent read errors.

Taking a look at the `__cache_spec` method, it seems like the writer uses its own manager rather than the source file's manager: https://github.com/hdmf-dev/hdmf-zarr/blob/5e48ab5e04512c5f5151d0d372eb2284f85dfeaa/src/hdmf_zarr/backend.py#L284

which is confirmed in a debugger. Setting a breakpoint at the start of that function:

```python
(Pdb) >? self.manager
<hdmf.build.manager.BuildManager object at 0x12cd55490>
(Pdb) >? self.manager.namespace_catalog.namespaces
('hdmf-common', 'hdmf-experimental', 'core')
```

and then going up to the initial call which has the source IO object:

```python
(Pdb) >? read_io.manager
<hdmf.build.manager.BuildManager object at 0x11029d610>
(Pdb) >? read_io.manager.namespace_catalog.namespaces
('hdmf-common', 'hdmf-experimental', 'core', 'ndx-aibs-ecephys')
```

If I do this just before calling `__cache_spec`:

```python
(Pdb) >? self.manager._BuildManager__type_map._TypeMap__ns_catalog = src_io.manager.namespace_catalog
```

then the namespace is correctly written, but the read still errors. The namespace is listed in `namespaces` here: https://github.com/hdmf-dev/hdmf-zarr/blob/5e48ab5e04512c5f5151d0d372eb2284f85dfeaa/src/hdmf_zarr/backend.py#L207

but when it reaches `BuildManager.load_namespaces`, the `deps` are empty. That's as far as i'm gonna make it tonight.

### Steps to Reproduce

```shell
from pynwb import NWBHDF5IO
from hdmf_zarr.nwb import NWBZarrIO

in_file = "/location/of/data/nwb/sub-738651046_ses-760693773.nwb"
out_file = "./test.zarr"

with NWBHDF5IO(in_file, 'r', load_namespaces=True) as read_io:
    with NWBZarrIO(out_file, mode="w") as export_io:
        export_io.export(src_io=read_io, write_args={'link_data':False})

[...]

PR #146, opened 2023-12-08
Write references in compound datasets as an array

### LINKED ISSUE #144: [Bug]: Iterative write for compound reference datasets takes 5ever
### What happened?

hello hello! checking out how you're structuring the zarr export so I can mimic it, and first try wasn't able to complete an export because after ~10m or so i had only written ~6MB of ~3GB and so i got to profiling. Searched for an issue but didn't find one, sorry if this has been raised already.

The problem is here: 
https://github.com/hdmf-dev/hdmf-zarr/blob/9c3b4bee013c825e2cdaccbe443009f88302803e/src/hdmf_zarr/backend.py#L1021-L1026

where each row has to be separately serialized on write. 

I see this behavior elsewhere like in `__list_fill__` so it seems like this is a general problem with write indexing into object-type zarr arrays - when I tried to collect all the new items in a list and write them like `dset[...] = new_items`, i got a shape error (and couldn't fix it by setting the shape correctly in the `require_dataset` ).

What did work, though, is doing this:

```python
if len(refs) > 0:
    dset = parent.require_dataset(name,
                  shape=(len(data), len(data[0])),
                  dtype=object,
                  object_codec=self.__codec_cls(),
                  **options['io_settings'])
    self._written_builders.set_written(builder)  # record that the builder has been written
    
    new_items = []
    for j, item in enumerate(data):
        new_item = list(item)
        for i in refs:
            new_item[i] = self.__get_ref(item[i], export_source=export_source)
        new_items.append(new_item)

    dset[...] = np.array(new_items)
```

The only difference that I could see is that rather than returning a python list when accessing the data again, it returned an (object-typed) numpy array.

The difference in timing is considerable (in my case, tool-breaking, which is why i'm raising as a bug rather than a feature request). When converting [this dataset](https://api.dandiarchive.org/api/dandisets/000021/versions/draft/assets/4513d0c9-1e2b-4c8a-aa22-ae81822537c9/) with 9 compound datasets w/ references, putting a timer around the above block:

| Array write | Iterative Write |
| ----------- | -------------- |
| 0.03479409217834473 | 1.675 |
| 0.010181903839111328 | 0.205 |
| 0.19667983055114746 | 62.746 |
| 1.0213229656219482 | 1808.36 |
| 2.000258207321167 | (dnf, it's been 30m and i need to go home) |
| 0.31616997718811035 | ... |
| 0.0024509429931640625 | ... |
| 0.3125300407409668 | ... |
| 0.01106715202331543 | ... |

I only spent ~20m or so profiling and debugging, so it doesn't replicate current top-level behavior exactly (creates an object array rather than a list of lists), but imo it might be better: the result is the same shape as the source dataset (so, in this case, an n x 3 array) and can be indexed as such, where the current implementation can't index by column since the array is just a one-dimensional array of serialized python lists. It also looks like it would be possible to improve it further still by using a numpy structured dtype (zarr seems to be able to handle that, at least from the docs, but the `require_dataset` method doesn't like it), etc. but hopefully useful as a first pass.

The "steps to reproduce" below is just what I put in a test script and ran `cProfile` on to diagnose:

### Steps to Reproduce

```shell
from pynwb import NWBHDF5IO

in_file = "/location/of/data/nwb/sub-738651046_ses-760693773.nwb"
out_file = "./test2.zarr"

with NWBHDF5IO(in_file, 'r', load_namespaces=True) as read_io:
    with NWBZarrIO(out_file, mode="w") as export_io:
        export_io.export(src_io=read_io, write_args={'link_data':False})
```

### Traceback

```shell
(not really a traceback kinda problem)
```

### Operating System

macOS

### Python Executable

Python

### Python Version

3.11

### Package Versions

[environment_for_issue.txt](https://github.com/hdmf-dev/hdmf-zarr/files/13596070/environment_for_issue.txt)

### Code of Conduct

sneakers-the-rat:
## Motivation

Fix https://github.com/hdmf-dev/hdmf-zarr/issues/144

Writes references in compound datasets as an same-sized array, rather than iteratively as an array of lists

## How to test the behavior?

Convert a dataset with compound datasets, confirm that it doesn't take a really long time.

Making a more detailed example blocked for me by https://github.com/hdmf-dev/hdmf-zarr/issues/145 bc i am not sure if i should actually work on this since it seems like devs are just doing their own version here (???) https://github.com/hdmf-dev/hdmf-zarr/pull/149

## Checklist

very tired and so leaving this as a draft for now


src/hdmf_zarr/backend.py
```diff
@@ -1012,17 +1012,20 @@ def write_dataset(self, **kwargs):  # noqa: C901
 
             if len(refs) > 0:
                 dset = parent.require_dataset(name,
-                                              shape=(len(data), ),
+                                              shape=(len(data), len(data[0])),
```
oruebel: This is not quite correct, because it expands the compound type, i.e., instead of an array of compound values, this represents the parts of the compound type as a new array dimension. I believe this is because currently each compound data value is described as a list, which numpy when you call `np.array(new_items)` by default interprets as a dimension, but instead I think we'd probably need to represent this a numpy array with a [structured dtype](https://numpy.org/doc/stable/user/basics.rec.html).

sneakers-the-rat: ah gotcha, my bad, will fix. i forgot that compound datasets are not just tables even tho they sorta look like them.

sneakers-the-rat:
Implemented writing as a compound dset, I used the information from the `options` passed to the function rather than using info from the `builder` class to reduce coupling, but if y'all prefer it that way it's a simple drop-in fix. 

So just after the dset is written...

```python
>>> (Pdb) dset[0]
(3798, 1, {'source': '.', 'path': '/processing/stimulus/timestamps', 'object_id': 'ff28f35d-9211-4bdb-b026-ad27d46367a3', 'source_object_id': 'a695c6dd-5c46-428d-a927-5c2027c4b653'})
>>> (Pdb) dset.dtype
dtype([('idx_start', '<i4'), ('count', '<i4'), ('timeseries', 'O')])
```

This version also doesn't have the same problem as https://github.com/hdmf-dev/hdmf-zarr/pull/149 , since the dataset itself is given the dtype, the returned type is a numpy void type rather than a tuple:

```python
>>> (Pdb) type(dset[0])
<class 'numpy.void'>
>>> (Pdb) dset[0][1] = 5
>>> (Pdb) # (it works)
```

whereas if the zarr dataset is created with `dtype=object`...

```python
>>> (Pdb) type(dset[0])
<class 'tuple'>
>>> (Pdb) dset[0][1] = 5
*** TypeError: 'tuple' object does not support item assignment
```

all tests pass. not sure what tests i would need to add here for this specific behavior but lmk

mavaylon1:
@sneakers-the-rat I'll take a look at the code coverage tonight, but from a quick glance it is passing.

mavaylon1:
@sneakers-the-rat Looks good to me. I can update the changelog for you if you'd like. After that, I'll do one last review and approve it.

sneakers-the-rat:
@mavaylon1 why don't you handle the changelog, since there's probably a way y'all like to write it/organize it with whatever other PRs have been going on, etc. :) 

lemme try to write a test, even if it's just as simple as testing that the write ends up in the right format (idk seems like a test for how long the operation took would be pretty flaky)

mavaylon1:
> @mavaylon1 why don't you handle the changelog, since there's probably a way y'all like to write it/organize it with whatever other PRs have been going on, etc. :) 
> 
> 
> 
> lemme try to write a test, even if it's just as simple as testing that the write ends up in the right format (idk seems like a test for how long the operation took would be pretty flaky)

Looking at the coverage, no further test is needed.

sneakers-the-rat:
Added a quick test that the array was written as a compound array, lmk if that's not the right way to test that - hardcoded values linked to the test dataset values, but seemed like the best way to differentiate that v. tuple behavior given the way the rest of the test is written. it also tests both the special cases for storage of objects and strings with `O` and `U25`

edit omg reading the rest of the tests i think i wrote this in the wrong place one sec
edit2: (moved to comment below)

mavaylon1:
> Added a quick test that the array was written as a compound array, lmk if that's not the right way to test that - hardcoded values linked to the test dataset values, but seemed like the best way to differentiate that v. tuple behavior given the way the rest of the test is written. it also tests both the special cases for storage of objects and strings with `O` and `U25`
> 
> 
> 
> edit omg reading the rest of the tests i think i wrote this in the wrong place one sec

So this should already be tested with existing code, that's why it is passing coverage prior.

sneakers-the-rat:
moving this out of edit so it stays in sequence:
 ok i don't see a better place for this, but can revert if not wanted. coverage should we the same since all the legs of the changed code are reached in the existing tests, but i don't see an explicit test for the dtype of the representation in the stored zarr array, so figured i would at least offer a test of the thing that i changed :)

rly:
@sneakers-the-rat Thanks for the updates and test!

> Looking at the coverage, no further test is needed.

@mavaylon1 Code coverage measures whether every line of the codebase has been run by the test suite. It doesn't reflect whether the codebase does what it is supposed to do and whether the new code correctly fixes the issue. In general, when an issue is found, a test (i.e., an `assert`) should be added that reproduces the issue and fails in the current main branch. The new code in the PR should make that test go from failing to passing. So here, even though all of the new code was touched by the test suite, a test should still be added to make sure the issue has actually been fixed and also make sure that any future changes to the codebase will not cause the same issue to arise.

sneakers-the-rat:
In this case the test also revealed a failure on windows, which is fun. Lemme debug

mavaylon1:
> @sneakers-the-rat Thanks for the updates and test!
> 
> 
> 
> > Looking at the coverage, no further test is needed.
> 
> 
> 
> @mavaylon1 Code coverage measures whether every line of the codebase has been run by the test suite. It doesn't reflect whether the codebase does what it is supposed to do and whether the new code correctly fixes the issue. In general, when an issue is found, a test (i.e., an `assert`) should be added that reproduces the issue and fails in the current main branch. The new code in the PR should make that test go from failing to passing. So here, even though all of the new code was touched by the test suite, a test should still be added to make sure the issue has actually been fixed and also make sure that any future changes to the codebase will not cause the same issue to arise.
> 
> 

@rly I'm aware. My point is the fact that there already should be a test that checks compound data is written as compound data. If not, then I understand adding one. The changes here, in practice, shouldn't break any tests because the results should be the same. Checking if the data was written as a compound data correctly should be the same of the tests that already check compound data (again these should already exist). From my understanding, the issue isn't a big but an optimization problem. The "fix" should be already covered by existing tests.

sneakers-the-rat:
From what i could tell, the checks that exist do pass, but they don't check explicitly for storage format. eg. the adjoining tests https://github.com/sneakers-the-rat/hdmf-zarr/blob/63922aea2339f64b96803394217498b833afc709/tests/unit/base_tests_zarrio.py#L444
that do `self.assertEqual(v[0], builder['data'][i][0])` would work for both compound dtypes and for object type arrays that store a 1D array of tuples since the indexing is the same in both cases. 

obviously y'all have a better sense of the tests that exist than I do, so if there already is one then great! I don't really care about the inclusion of the one crappy lil test I added (or even really the storage format), my main goal here is to just help out making writing happen at a reasonable speed, so lmk what's good <3

edit: sorry since there is a failing test - i don't have a windows machine to debug, but i can just spam `warning`s as print statements in the GH actions until it works lol. if y'all have windows machines and debug before i can get to it, then u win the debugging race!

mavaylon1:
> From what i could tell, the checks that exist do pass, but they don't check explicitly for storage format. eg. the adjoining tests https://github.com/sneakers-the-rat/hdmf-zarr/blob/63922aea2339f64b96803394217498b833afc709/tests/unit/base_tests_zarrio.py#L444
> 
> that do `self.assertEqual(v[0], builder['data'][i][0])` would work for both compound dtypes and for object type arrays that store a 1D array of tuples since the indexing is the same in both cases. 
> 
> 
> 
> obviously y'all have a better sense of the tests that exist than I do, so if there already is one then great! I don't really care about the inclusion of the one crappy lil test I added (or even really the storage format), my main goal here is to just help out making writing happen at a reasonable speed, so lmk what's good <3
> 
> 
> 
> edit: sorry since there is a failing test - i don't have a windows machine to debug, but i can just spam `warning`s as print statements in the GH actions until it works lol. if y'all have windows machines and debug before i can get to it, then u win the debugging race!

I'll look after the break. By all means if there's a gap in the test we should fill it. My point was based on the assumption that there's tests that already do what we want because this isn't a format change. Just an order of operations. But if this process "exposed" a gap, then let's fill it.

mavaylon1:
@sneakers-the-rat I updated the test to our usual use of TestCase "assertEqual". This gave more insight on the issue for windows: `AssertionError: dtype([('id', '<i4'), ('name', '<U25'), ('reference', 'O')]) != dtype([('id', '<i8'), ('name', '<U25'), ('reference', 'O')])`

The id types are not matching.

sneakers-the-rat:
ope. so windows treats `np.dtype('int')` as 32-bits by default i guess, good to know.

looking to see if there was an explicit type map, i can see that there is and that I missed it. That also shows that my check for `is str` was too strict, since presumably all strings should be saved as `'U25'` (I'm just basing that off what you specified here https://github.com/hdmf-dev/hdmf-zarr/blob/005bfbc0eb3ecee884170ec2a93b7d20c63c9655/src/hdmf_zarr/backend.py#L1032 which seems like it could truncate strings to me, but that part ofc is up to y'all). Replaced it with a check for all types that are mapped to `str` in `__dtypes`.

It seems like there should be a single string dtype -> numpy dtype map method, but the `__resolve_dtype_helper__` method has a problem in that zarr doesn't support variable length unicode strings ( https://zarr.readthedocs.io/en/stable/tutorial.html#string-arrays ), and so the mapping is `str -> '<U'`, which zarr silently drops and when read it returns an empty string. I kept most of the existing type creation stuff, but added a call to `__resolve_dtype_helper__` in the else case, and added a failure mode for an unhandled `else` in that method since the existing `else` implicitly assumed that the dtype was a list.

Feels like the dtype conversion logic is getting pretty dispersed at this point, would be nice to consolidate that into one or a few maps (seems like would be good to have separate ones for intended python type vs. storage type, eg. `str` vs `'<U25'`, but i won't turn this PR into a big refactor <3

---

This might reveal another issue, but i'm not sure - just prior to the changes I made, a `type_str` object is made that makes use of the type map: https://github.com/hdmf-dev/hdmf-zarr/blob/9692a98f0568e0f968c89411a67bd4f483c91d9d/src/hdmf_zarr/backend.py#L1010

That gets saved as a dataset attribute: https://github.com/hdmf-dev/hdmf-zarr/blob/9692a98f0568e0f968c89411a67bd4f483c91d9d/src/hdmf_zarr/backend.py#L1020

So if that is supposed to be used by the reader to re-hydrate the object and cast to correct types, it isn't being used. But i'm not sure that's what that is for, so idk.

mavaylon1:
> ope. so windows treats `np.dtype('int')` as 32-bits by default i guess, good to know.
> 
> 
> 
> looking to see if there was an explicit type map, i can see that there is and that I missed it. That also shows that my check for `is str` was too strict, since presumably all strings should be saved as `'U25'` (I'm just basing that off what you specified here https://github.com/hdmf-dev/hdmf-zarr/blob/005bfbc0eb3ecee884170ec2a93b7d20c63c9655/src/hdmf_zarr/backend.py#L1032 which seems like it could truncate strings to me, but that part ofc is up to y'all). Replaced it with a check for all types that are mapped to `str` in `__dtypes`.
> 
> 
> 
> It seems like there should be a single string dtype -> numpy dtype map method, but the `__resolve_dtype_helper__` method has a problem in that zarr doesn't support variable length unicode strings ( https://zarr.readthedocs.io/en/stable/tutorial.html#string-arrays ), and so the mapping is `str -> '<U'`, which zarr silently drops and when read it returns an empty string. I kept most of the existing type creation stuff, but added a call to `__resolve_dtype_helper__` in the else case, and added a failure mode for an unhandled `else` in that method since the existing `else` implicitly assumed that the dtype was a list.
> 
> 
> 
> Feels like the dtype conversion logic is getting pretty dispersed at this point, would be nice to consolidate that into one or a few maps (seems like would be good to have separate ones for intended python type vs. storage type, eg. `str` vs `'<U25'`, but i won't turn this PR into a big refactor <3
> 
> 
> 
> ---
> 
> 
> 
> This might reveal another issue, but i'm not sure - just prior to the changes I made, a `type_str` object is made that makes use of the type map: https://github.com/hdmf-dev/hdmf-zarr/blob/9692a98f0568e0f968c89411a67bd4f483c91d9d/src/hdmf_zarr/backend.py#L1010
> 
> 
> 
> That gets saved as a dataset attribute: https://github.com/hdmf-dev/hdmf-zarr/blob/9692a98f0568e0f968c89411a67bd4f483c91d9d/src/hdmf_zarr/backend.py#L1020
> 
> 
> 
> So if that is supposed to be used by the reader to re-hydrate the object and cast to correct types, it isn't being used. But i'm not sure that's what that is for, so idk.
> 
> 
> 
> 

The strings don't need to be length 25. I also found that doing <U would return empty strings downstream. The U25 was me testing to see if providing a cap would fix that and it does. That's an open ended discussion that I thought was fixed with your implementation, but I think there's something lingering that has yet to be solidified. I'm going to dive into this Friday and see what I can come up with.

sneakers-the-rat:
Basically all that needs to happen is to decide on either a single length, or calculate the Max needed length from the data, ideally in a single mapping method that can get hooked into from the various places one would need to get a zarr storage dtype from a source dtype. Hopefully not too complicated :)

src/hdmf_zarr/backend.py
```diff
@@ -1011,18 +1011,49 @@ def write_dataset(self, **kwargs):  # noqa: C901
...
+                # Create dtype for storage, replacing values to match hdmf's hdf5 behavior
+                # ---
+                # TODO: Replace with a simple one-liner once __resolve_dtype_helper__ is
+                # compatible with zarr's need for fixed-length string dtypes.
+                # dtype = self.__resolve_dtype_helper__(options['dtype'])
+
+                new_dtype = []
+                for field in options['dtype']:
+                    if field['dtype'] is str or field['dtype'] in (
+                            'str', 'text', 'utf', 'utf8', 'utf-8', 'isodatetime'
```
mavaylon1, line 1040: I'm actually not sure if this level of specificity is needed. @oruebel thoughts?

mavaylon1: These are listed in the dict and mapped to 'str' in line 1114

src/hdmf_zarr/backend.py
```diff
@@ -1011,18 +1011,49 @@ def write_dataset(self, **kwargs):  # noqa: C901
...
+                    ):
+                        new_dtype.append((field['name'], 'U25'))
```
mavaylon1: @oruebel I feel that setting a cap on the length, though quick and it works, is not general. I think to get this ticket wrapped up we decide on a temporary cap and investigate why on read the field turns to an empty string if 'U25' is just 'U'. Thoughts?

oruebel: I'm not sure if Zarr support variable length structured data types. That's something to look at. If not, then we may need to either: 1) introspect the data to determine the string length or 2) maybe it's possible to use 'O' as instead of 'U25' in the compound dtype  or 3) store variable length compound types as objects.  Did you try the second option. i.e., use 'O' or 'object' in the dtype instead of 'U'?

src/hdmf_zarr/backend.py
sneakers-the-rat: Right, im trying to mimic that dict, but `str` just get stored as a variable length string which zarr drops, as per https://github.com/hdmf-dev/hdmf-zarr/pull/146#issuecomment-1870039783

src/hdmf_zarr/backend.py
sneakers-the-rat: It doesnt support variable length strings, no. Object would almost certainly work at the cost of needing to invoke a json serializer, but since we already have to do that for the reference object it would probs be fine. Introspecting the string would also probably be fine but gives me fragility vibes. Can try both, would the deciding factor be perf or what?

oruebel: > Introspecting the string would also probably be fine but gives me fragility vibes

I agree, introspecting on write is not desirable both from a fragility and performance perspective.

mavaylon1:
Hey @sneakers-the-rat let me know if you want to finish this up or if you want me to.

sneakers-the-rat:
Ah I was not sure what approach we were taking, I can do some tests today to finish it out

sneakers-the-rat:
Alright, I didn't do seriously careful profiling here, but switching out the `'U25'` string with `'O'` is a perf wash, and when read back seems to be seamlessly recast back to a string.

On a single run (not robust) of the relevant test cases (cProfile's cumulative time):
| Case | U25 | Object |
| ----- | ---- | ------- |
| `test_read_reference_compound_buf` | 0.0375 | 0.0455 |
| `test_read_reference_compound` | 0.0356 | 0.0325 |
| `test_write_reference_compound` | 0.0381 | 0.0388 | 
|  total | 0.0911 | 0.1 | 

But on a "naturalistic" case where I just timed the whole write of the dataset i had referenced in the issue, oddly the object encoding was faster. for the `write_dataset` method:
- object: 38s
- U25: 39.1s

So pretty much the same. I won't add a test case for very long strings since if I did it would just be a hackjob attached to the presently modified test methods, and that seems like something better suited for a text fixture/parameterization.

sneakers-the-rat:
Sorry for the force push, just to amend an incorrect commit message

sneakers-the-rat:
I think this is good to go at this point :) lmk if changes still needed

mavaylon1 (approved):

PR #149, opened 2023-12-13
Write references in compound datasets at the end

### LINKED ISSUE #144: [Bug]: Iterative write for compound reference datasets takes 5ever

mavaylon1:
## Motivation

Fix #144 

Notes:
The fix at face value is the following:
```
for j, item in enumerate(data):
        new_item = list(item)
        for i in refs:
             new_item[i] = self.__get_ref(item[i], export_source=export_source)
        new_items.append(tuple(new_item))
        
        generated_dtype = []
        for item in builder.dtype:
              if item['dtype'] == type('string'):
                  item['dtype'] = 'U25'
              generated_dtype.append((item['name'], item['dtype']))

 dset[...] = np.array(new_items, dtype=generated_dtype)
```

TLDR: We want to write outside of the the loop. What does this lead to?

To keep our compound dataset in its original form (i.e not expanded) we can use a structured array an pull the dtypes from the builder. However, it is required that the "data" in the array be in tuples and not lists (I believe it can also be in np.arrays but I have not tested this). 

Having these as tuples makes things a bit hard when retrieving data on read. For example, in the test `test_read_reference_compound`
```
def test_read_reference_compound(self):
        self.test_write_reference_compound()
        self.read()
        builder = self.createReferenceCompoundBuilder()['ref_dataset']
        read_builder = self.root['ref_dataset']
```
If I try to run `read_builder['data'][0]` I will get an error that `*** TypeError: 'tuple' object does not support item assignment`. Digging deeper it is because our `get_item` swaps/will reassign values in the tuple. This is not allowed because it is a tuple. Now I can convert that to a list on `_get_item_` and that does work. However, it is a little hacky. It also doesn't cover when my index is ":" vs a single integer. 

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/src/hdmf_zarr/backend.py
+++ b/src/hdmf_zarr/backend.py
@@ -1018,11 +1018,20 @@ class ZarrIO(HDMFIO):
                                               **options['io_settings'])
                 self._written_builders.set_written(builder)  # record that the builder has been written
                 dset.attrs['zarr_dtype'] = type_str
+                new_items = []
                 for j, item in enumerate(data):
                     new_item = list(item)
                     for i in refs:
                         new_item[i] = self.__get_ref(item[i], export_source=export_source)
-                    dset[j] = new_item
+                    new_items.append(tuple(new_item))
+                    # dset[j] = new_item
+
+                generated_dtype = []
+                for item in builder.dtype:
+                    if item['dtype'] == type('string'):
+                        item['dtype'] = 'U25'
+                    generated_dtype.append((item['name'], item['dtype']))
+                dset[...] = np.array(new_items, dtype=generated_dtype)
             else:
                 # write a compound datatype
                 dset = self.__list_fill__(parent, name, data, options)
--- a/src/hdmf_zarr/backend.py
+++ b/src/hdmf_zarr/backend.py
@@ -1011,27 +1011,59 @@ class ZarrIO(HDMFIO):
                     type_str.append(self.__serial_dtype__(t)[0])
 
             if len(refs) > 0:
-                dset = parent.require_dataset(name,
-                                              shape=(len(data), ),
-                                              dtype=object,
-                                              object_codec=self.__codec_cls(),
-                                              **options['io_settings'])
+                # dset = parent.require_dataset(name,
+                #                               shape=(len(data), ),
+                #                               dtype=object,
+                #                               object_codec=self.__codec_cls(),
+                #                               **options['io_settings'])
                 self._written_builders.set_written(builder)  # record that the builder has been written
-                dset.attrs['zarr_dtype'] = type_str
+                # dset.attrs['zarr_dtype'] = type_str
                 new_items = []
                 for j, item in enumerate(data):
                     new_item = list(item)
                     for i in refs:
                         new_item[i] = self.__get_ref(item[i], export_source=export_source)
                     new_items.append(tuple(new_item))
-                    # dset[j] = new_item
 
-                generated_dtype = []
-                for item in builder.dtype:
-                    if item['dtype'] == type('string'):
-                        item['dtype'] = 'U25'
-                    generated_dtype.append((item['name'], item['dtype']))
-                dset[...] = np.array(new_items, dtype=generated_dtype)
+                # Create dtype for storage, replacing values to match hdmf's hdf5 behavior
+                # ---
+                # TODO: Replace with a simple one-liner once __resolve_dtype_helper__ is
+                # compatible with zarr's need for fixed-length string dtypes.
+                # dtype = self.__resolve_dtype_helper__(options['dtype'])
+
+                new_dtype = []
+                for field in options['dtype']:
+                    if field['dtype'] is str or field['dtype'] in ('str', 'text', 'utf', 'utf8', 'utf-8', 'isodatetime'):
+                        new_dtype.append((field['name'], 'U'))
+                    elif isinstance(field['dtype'], dict):
+                        # eg. for some references, dtype will be of the form
+                        # {'target_type': 'Baz', 'reftype': 'object'}
+                        # which should just get serialized as an object
+                        new_dtype.append((field['name'], 'O'))
+                    else:
+                        new_dtype.append((field['name'], self.__resolve_dtype_helper__(field['dtype'])))
+                dtype = np.dtype(new_dtype)
+
+                # cast and store compound dataset
+                arr = np.array(new_items, dtype=dtype)
+                dset = parent.require_dataset(
+                    name,
+                    shape=(len(arr),),
+                    dtype=dtype,
+                    object_codec=self.__codec_cls(),
+                    **options['io_settings']
+                )
+                dset.attrs['zarr_dtype'] = type_str
+                dset[...] = arr
+
+
+
+                # generated_dtype = []
+                # for item in builder.dtype:
+                #     if item['dtype'] == type('string'):
+                #         item['dtype'] = 'U25'
+                #     generated_dtype.append((item['name'], item['dtype']))
+                # dset[...] = np.array(new_items, dtype=generated_dtype)
             else:
                 # write a compound datatype
                 dset = self.__list_fill__(parent, name, data, options)
@@ -1156,8 +1188,10 @@ class ZarrIO(HDMFIO):
             return cls.__dtypes.get(dtype)
         elif isinstance(dtype, dict):
             return cls.__dtypes.get(dtype['reftype'])
-        else:
+        elif isinstance(dtype, list):
             return np.dtype([(x['name'], cls.__resolve_dtype_helper__(x['dtype'])) for x in dtype])
+        else:
+            raise ValueError(f'Cant resolve dtype {dtype}')
 
     @classmethod
     def get_type(cls, data):
--- a/src/hdmf_zarr/zarr_utils.py
+++ b/src/hdmf_zarr/zarr_utils.py
@@ -151,7 +151,8 @@ class AbstractZarrTableDataset(DatasetOfReferences):
         return self.__dtype
 
     def __getitem__(self, arg):
-        rows = copy(super().__getitem__(arg))
+        rows = list(copy(super().__getitem__(arg)))
+        # breakpoint()
         if np.issubdtype(type(arg), np.integer):
             self.__swap_refs(rows)
         else:
--- a/tests/unit/base_tests_zarrio.py
+++ b/tests/unit/base_tests_zarrio.py
@@ -426,7 +426,7 @@ class BaseTestZarrWriter(BaseZarrWriterTestCase):
         builder = self.createReferenceBuilder()['ref_dataset']
         read_builder = self.root['ref_dataset']
         # Load the linked arrays and confirm we get the same data as we had in the original builder
-        breakpoint()
+        # breakpoint()
         # for i, v in enumerate(read_builder['data']):
         #     self.assertTrue(np.all(builder['data'][i]['builder']['data'] == v['data'][:]))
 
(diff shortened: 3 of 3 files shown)


mavaylon1:
@oruebel As I was going through the tests, I saw that the way I had to interact with the read data differed from before. Diving deeper it comes down to what I said above where we are now using tuples, which cannot change. The fix (even if it is a little hacky) is to convert to a list when trying to use index notation to "get". I wanted your thoughts on the matter and I also think this may need to go all the way up the parent hierarchy to HDMFDataset.

oruebel:
> Digging deeper it is because our `get_item` swaps/will reassign values in the tuple.

Could you please point me to the lines of code where the tuple is being modified on read so I can take a look to better understand what is going. I'm not sure right now why the tuples are being modified on read (maybe to resolve the references).

mavaylon1:
> > Digging deeper it is because our `get_item` swaps/will reassign values in the tuple.
> 
> Could you please point me to the lines of code where the tuple is being modified on read so I can take a look to better understand what is going. I'm not sure right now why the tuples are being modified on read (maybe to resolve the references).

[Lines ](https://github.com/hdmf-dev/hdmf-zarr/blob/9692a98f0568e0f968c89411a67bd4f483c91d9d/src/hdmf_zarr/zarr_utils.py#L153-L165)

src/hdmf_zarr/zarr_utils.py
```diff
@@ -151,7 +151,8 @@ def dtype(self):
         return self.__dtype
 
     def __getitem__(self, arg):
-        rows = copy(super().__getitem__(arg))
+        rows = list(copy(super().__getitem__(arg)))
+        # breakpoint()
```
oruebel, line 155: If I understand what you are saying correctly, the issue appears when ``self.__swap_refs`` is being called because the individual rows are now represented as tuples. I.e, `rows` is either a tuple (if `arg` is an int) or a list of tuples (if `arg` selects multiple rows). If that is the issue, then I think the we could update ``self.__swap_refs`` to return a new instance of the row (rather than updating the existing one). Here is what I think might work: 


```
def __getitem__(self, arg):
        rows = copy(super().__getitem__(arg))
        if np.issubdtype(type(arg), np.integer):
            rows =  self.__swap_refs(rows)
        else:
            rows = [self.__swap_refs(row) for row in rows]
       return rows

def __swap_refs(self, row):
        updated_row = list(row)  # convert tuple to a list so we can update it
        for i in self.__refgetters:
            getref = self.__refgetters[i]
            updated_row[i] = getref(row[i])
        return updated_row   # TODO if we want to keep these as tuples then we could convert them back here 
```

sneakers-the-rat:
I was just going to work on https://github.com/hdmf-dev/hdmf-zarr/pull/146 and saw this parallel PR, should I work on this instead or what

src/hdmf_zarr/backend.py
```diff
@@ -1018,11 +1018,20 @@ def write_dataset(self, **kwargs):  # noqa: C901
...
                     for i in refs:
                         new_item[i] = self.__get_ref(item[i], export_source=export_source)
-                    dset[j] = new_item
+                    new_items.append(tuple(new_item))
+                    # dset[j] = new_item
+
+                generated_dtype = []
+                for item in builder.dtype:
+                    if item['dtype'] == type('string'):
+                        item['dtype'] = 'U25'
```
sneakers-the-rat: this mutates the builder's dtype right?

sneakers-the-rat:
Wrote this here: https://github.com/hdmf-dev/hdmf-zarr/pull/146#issuecomment-1855049073

but that tuple problem is just from not setting the dtype in the zarr dataset, so it ignores the dtype you give to `np.array` and just stores as tuples. If you correctly set that then it behaves just fine.

So do this:
https://github.com/hdmf-dev/hdmf-zarr/blob/8100341512324af25718ab93dc7fc54b68f03469/src/hdmf_zarr/backend.py#L1044

instead of this:
https://github.com/hdmf-dev/hdmf-zarr/blob/005bfbc0eb3ecee884170ec2a93b7d20c63c9655/src/hdmf_zarr/backend.py#L1016

src/hdmf_zarr/backend.py
mavaylon1: Yes. When numpy gets the "str class" it removes the actual value and on read returns an empty string.

mavaylon1:
> I was just going to work on #146 and saw this parallel PR, should I work on this instead or what

Yes please

mavaylon1:
> read_builder['data'][0]

Interesting. Could you explain more?

mavaylon1:
> Wrote this here: [#146 (comment)](https://github.com/hdmf-dev/hdmf-zarr/pull/146#issuecomment-1855049073)
> 
> but that tuple problem is just from not setting the dtype in the zarr dataset, so it ignores the dtype you give to `np.array` and just stores as tuples. If you correctly set that then it behaves just fine.
> 
> So do this:
> 
> https://github.com/hdmf-dev/hdmf-zarr/blob/8100341512324af25718ab93dc7fc54b68f03469/src/hdmf_zarr/backend.py#L1044
> 
> instead of this:
> 
> https://github.com/hdmf-dev/hdmf-zarr/blob/005bfbc0eb3ecee884170ec2a93b7d20c63c9655/src/hdmf_zarr/backend.py#L1016

Oh it stores it as <class 'numpy.void'> when you provide a dtype instead of a tuple.

src/hdmf_zarr/backend.py
sneakers-the-rat: not sure what you mean, when numpy gets `str` where? What i mean is that this would mutate the `builder.dtype` dict, which is almost always undesirable, especially since the builder is passed in as an arg, mutating your args will buy you invisible bugs later.

eg.

```python
>>> mydtypes = [{'dtype': 'int', 'name': 'test'}, {'dtype': 'float', 'name': 'test2'}]
>>> for item in mydtypes:
...    if item['dtype'] == 'int':
...        item['dtype'] = 'newvalue'
>>> mydtypes
[{'dtype': 'newvalue', 'name': 'test'}, {'dtype': 'float', 'name': 'test2'}]
```

instead do 

```python
for item in builder.dtype:
    if item['dtype'] is str:
        generated_dtype.append((item['name'], 'U25'))
    else:
        generated_dtype.append((item['name'], item['dtype']))
```

which doesn't mutate `builder.dtype`. see https://github.com/hdmf-dev/hdmf-zarr/blob/8100341512324af25718ab93dc7fc54b68f03469/src/hdmf_zarr/backend.py#L1028-L1036

sneakers-the-rat:
> Yes please

Alright, lmk how - i don't have commit permissions on this repo so can't modify this branch. Usually bad etiquette to take the several hours of work i've done here and just rewrite it without credit, but whatevs.

> Oh it stores it as <class 'numpy.void'> when you provide a dtype instead of a tuple.

right, and that's the desired behavior for storing a structured array, which is what I thought the goal was. As this PR is written, the constructed `dtype` is just ignored and the array is just stored as serialized tuples.

src/hdmf_zarr/backend.py
mavaylon1: I see what you did.

mavaylon1:
> > Yes please
> 
> 
> 
> Alright, lmk how - i don't have commit permissions on this repo so can't modify this branch. Usually bad etiquette to take the several hours of work i've done here and just rewrite it without credit, but whatevs.
> 
> 
> 
> > Oh it stores it as <class 'numpy.void'> when you provide a dtype instead of a tuple.
> 
> 
> 
> right, and that's the desired behavior for storing a structured array, which is what I thought the goal was. As this PR is written, the constructed `dtype` is just ignored and the array is just stored as serialized tuples.

I believe we were working on this in parallel. We can just keep it to your branch on your fork to give you credit. The intent was for me to dig around from the last checkpoint I saw on yours. Let me know when you're done and I'll review it. I'll continue to play around here for educational purposes.

sneakers-the-rat:
ya sorry i don't mean to be prickly but this has happened a number of times to me lol and usually i don't care, but when last looking for a job it actually was important to be able to demonstrate contribs. I think the patch is done except for tests and docs?

mavaylon1:
> ya sorry i don't mean to be prickly but this has happened a number of times to me lol and usually i don't care, but when last looking for a job it actually was important to be able to demonstrate contribs. I think the patch is done except for tests and docs?

I can check the tests, but I will be out of town starting tomorrow. I'll go through it will most likely finalize this for a merge by Tuesday.

sneakers-the-rat:
lmk what else needs to be written re: docs and tests, happy to do that! i try to be as labor-neutral as i can be when raising issues/PRs :). take ur time, no rush. also sorry for my tone earlier, frustrating day.

rly:
@mavaylon1 can this be closed, now that #146 has been merged?

PR #172, opened 2024-03-03
Update to support python 3.12

mavaylon1:
## Motivation

Support python 3.12

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -66,6 +66,7 @@ body:
         - "3.9"
         - "3.10"
         - "3.11"
+        - "3.12"
     validations:
       required: true
   - type: textarea
--- a/.github/workflows/HDMF_dev.yaml
+++ b/.github/workflows/HDMF_dev.yaml
@@ -16,10 +16,10 @@ jobs:
         with:
             fetch-depth: 0  # tags are required for versioneer to determine the version
       
-      - name: Set up Python 3.11
-        uses: actions/setup-python@v4
+      - name: Set up Python 3.12
+        uses: actions/setup-python@v5
         with:
-          python-version: "3.11"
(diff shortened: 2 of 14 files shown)


mavaylon1:
Note: After updating the .toml, tox, and workflows, I did a search for "3.11". I'll do another look to see what else needs to be bumped to 3.12

mavaylon1:
@oruebel The bump to 3.12 seems to have bequeathed some gallery errors that need to be resolved. I'd like to make it a separate ticket/PR to not hold this up. Thoughts? I need to do another pass through before requesting a review.

Update: I just ended up fixing it here. A new warning filter for test_gallery.py

mavaylon1:
Review Note: I did not update the changelog since this is not a user facing update.

rly:
> Review Note: I did not update the changelog since this is not a user facing update.

Please add a changelog -- The changelog is for any change to the software, and is for both users and developers, especially to check for changes that affect compatibility (here, for example, we announce that PyNWB is officially compatible with Python 3.12 - some downstream packages do not support Python 3.12 because pynwb does not so it's important for us to announce this). 

In contrast, when making changes to the schema repo, the schema release notes should not be updated when there are no changes to the schema files. We could add a changelog there for all changes to the repo, but its not really necessary because the main purpose of the schema repo is to hold the schema files. The surrounding code is for internal use.

rly:
I mean HDMF-Zarr, not PyNWB, though my example about downstream packages not supporting a version of Python because PyNWB does not officially support is is true for PyNWB, not HDMF-Zarr (AFAIK)

mavaylon1:
> I mean HDMF-Zarr, not PyNWB, though my example about downstream packages not supporting a version of Python because PyNWB does not officially support is is true for PyNWB, not HDMF-Zarr (AFAIK)

Got it. Done.

rly (approved):

PR #173, opened 2024-03-05
default object_codec_class -> JSON

### LINKED ISSUE #171: [Feature]: Write zarr without using pickle

I was working on adding nwb-zarr support to [neurosift](https://github.com/flatironinstitute/neurosift) and I ran into a problem where many of the datasets within the zarr were pickle-encoded. It is difficult to read pickled data in languages other than Python, and it is practically impossible using JavaScript in the browser. So I would like to request that hdmf-zarr be updated to avoid using pickling when writing datasets.

_No response_

@bendichter and I created a script for doing the h5->zarr conversion without using pickle:
https://gist.github.com/magland/39c5c5c5afabb982b5dd5d2367f2179b

and we confirmed that it (a) the output is loadable in hdmf-zarr and (b) it works with Neurosift

We have only tried it on [the one nwb file so far](https://hdmf-zarr.readthedocs.io/en/latest/tutorials/plot_convert_nwb_hdf5.html#sphx-glr-tutorials-plot-convert-nwb-hdf5-py).

Perhaps this script could be helpful in making the modifications.

I perused the hdmf-zarr source code, but it wasn't clear to me where the changes needed to take place.

Tagging: @rly 

Yes.

### Code of Conduct

magland:
## Motivation

See #171 

> I was working on adding nwb-zarr support to [neurosift](https://github.com/flatironinstitute/neurosift) and I ran into a problem where many of the datasets within the zarr were pickle-encoded. It is difficult to read pickled data in languages other than Python, and it is practically impossible using JavaScript in the browser. So I would like to request that hdmf-zarr be updated to avoid using pickling when writing datasets.

In this PR I adjusted the default object_codec_class to numcodecs.JSON.

In addition, I also exposed this parameter in the constructor of NWBZarrIO

Fix #171 

## How to test the behavior?

I ran the following with my forked version, and verified that the resulting zarr opens properly in Neurosift:

```python
import os
from pynwb import NWBHDF5IO
import numcodecs

filename = 'sub-anm00239123_ses-20170627T093549_ecephys+ogen.nwb'

def main():
    zarr_filename = 'test.zarr'
    with NWBHDF5IO(path=filename, mode='r', load_namespaces=False) as read_io:  # Create HDF5 IO object for read
        with NWBZarrIO(path=zarr_filename, mode='w') as export_io:         # Create Zarr IO object for write
            export_io.export(src_io=read_io, write_args=dict(link_data=False))   # Export from HDF5 to Zarr

if __name__ == "__main__":
    main()
```

## Checklist

Note: I wasn't sure how to update CHANGELOG.md -- I mean where to put the new tet

### DIFF
--- a/src/hdmf_zarr/backend.py
+++ b/src/hdmf_zarr/backend.py
@@ -1052,7 +1052,7 @@ class ZarrIO(HDMFIO):
 
                 object_codec = self.__codec_cls()
                 if not isinstance(object_codec, numcodecs.Pickle):
-                    print(f'Resorting to Pickle codec for dataset {name} of {parent.name}')
+                    warnings.warn(f'Resorting to Pickle codec for dataset {name} of {parent.name}')
                     object_codec = numcodecs.Pickle()
 
                 # cast and store compound dataset
@@ -1287,7 +1287,7 @@ class ZarrIO(HDMFIO):
         if has_structured_array:
             object_codec = io_settings.get('object_codec')
             if not isinstance(object_codec, numcodecs.Pickle):
-                print(f'Warning: Resorting to Pickle codec for {name} of {parent.name}.')
+                warnings.warn(f'Resorting to Pickle codec for {name} of {parent.name}.')
                 io_settings['object_codec'] = numcodecs.Pickle()
 
         # Create the dataset
--- a/src/hdmf_zarr/backend.py
+++ b/src/hdmf_zarr/backend.py
@@ -94,7 +94,7 @@ class ZarrIO(HDMFIO):
              'default': None},
             {'name': 'object_codec_class', 'type': None,
              'doc': 'Set the numcodec object codec class to be used to encode objects.'
-                    'Use numcodecs.pickles.Pickle by default.',
+                    'Use numcodecs.JSON by default.',
              'default': None},
             {'name': 'storage_options', 'type': dict,
              'doc': 'Zarr storage options to read remote folders',
--- a/src/hdmf_zarr/backend.py
+++ b/src/hdmf_zarr/backend.py
@@ -1050,13 +1050,18 @@ class ZarrIO(HDMFIO):
                         new_dtype.append((field['name'], self.__resolve_dtype_helper__(field['dtype'])))
                 dtype = np.dtype(new_dtype)
 
+                object_codec = self.__codec_cls()
+                if not isinstance(object_codec, numcodecs.Pickle):
+                    print(f'Resorting to Pickle codec for dataset {name} of {parent.name}')
+                    object_codec = numcodecs.Pickle()
+
                 # cast and store compound dataset
                 arr = np.array(new_items, dtype=dtype)
                 dset = parent.require_dataset(
                     name,
                     shape=(len(arr),),
                     dtype=dtype,
-                    object_codec=self.__codec_cls(),
+                    object_codec=object_codec,
                     **options['io_settings']
                 )
                 dset.attrs['zarr_dtype'] = type_str
@@ -1267,6 +1272,23 @@ class ZarrIO(HDMFIO):
         # Determine the shape from the data if all other cases have not been hit
         else:
             data_shape = get_data_shape(data)
+
+        # Let's check to see if we have a structured array somewhere in the data
+        # If we do, then we are going to resort to pickling the data and
+        # printing a warning.
+        has_structured_array = False
+        if dtype == object:
+            for c in np.ndindex(data_shape):
+                o = data
+                for i in c:
+                    o = o[i]
+                if isinstance(o, np.void) and o.dtype.names is not None:
+                    has_structured_array = True
+        if has_structured_array:
+            object_codec = io_settings.get('object_codec')
+            if not isinstance(object_codec, numcodecs.Pickle):
+                print(f'Warning: Resorting to Pickle codec for {name} of {parent.name}.')
+                io_settings['object_codec'] = numcodecs.Pickle()
 
         # Create the dataset
         dset = parent.require_dataset(name, shape=data_shape, dtype=dtype, **io_settings)
--- a/src/hdmf_zarr/backend.py
+++ b/src/hdmf_zarr/backend.py
@@ -120,8 +120,8 @@ class ZarrIO(HDMFIO):
         self.__built = dict()
         self._written_builders = WriteStatusTracker()  # track which builders were written (or read) by this IO object
         self.__dci_queue = None  # Will be initialized on call to io.write
-        # Codec class to be used. Alternates, e.g., =numcodecs.JSON
-        self.__codec_cls = numcodecs.pickles.Pickle if object_codec_class is None else object_codec_class
+        # Codec class to be used. Alternates, e.g., =numcodecs.pickles.Pickle
+        self.__codec_cls = numcodecs.JSON if object_codec_class is None else object_codec_class
         source_path = self.__path
         if isinstance(self.__path, SUPPORTED_ZARR_STORES):
             source_path = self.__path.path
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # HDMF-ZARR Changelog
 
+## 0.7.0 (Upcoming)
+
+### Enhancements
+
+* Changed default object_codec_class for ZarrIO to numcodecs.JSON. The issue with the old default (numcodecs.Pickle) was that it was not readable outside of Python. Also, exposed the object_codec_class as a parameter to the NWBZarrIO constructor.
+
 ## 0.6.0 (February 21, 2024)
 
 ### Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Enhancements
 
-* Changed default object_codec_class for ZarrIO to numcodecs.JSON. The issue with the old default (numcodecs.Pickle) was that it was not readable outside of Python. Also, exposed the object_codec_class as a parameter to the NWBZarrIO constructor.
+* Changed default object_codec_class for ZarrIO to numcodecs.JSON. The issue with the old default (numcodecs.Pickle) was that it was not readable outside of Python. Exposed the object_codec_class as a parameter to the NWBZarrIO constructor. Resort to Pickle for complex cases such as structured arrays or compound datasets with refs.
 
 ## 0.6.0 (February 21, 2024)
 
--- a/docs/source/storage.rst
+++ b/docs/source/storage.rst
@@ -249,7 +249,7 @@ Zarr file. The individual object references are defined in the
 :py:class:`~hdmf_zarr.backend.ZarrIO` as py:class:`~hdmf_zarr.utils.ZarrReference` object created via
 the :py:meth:`~hdmf_zarr.backend.ZarrIO.__get_ref` helper function.
 
-By default, :py:class:`~hdmf_zarr.backend.ZarrIO` uses the ``numcodecs.pickles.Pickle`` codec to
+By default, :py:class:`~hdmf_zarr.backend.ZarrIO` uses the ``numcodecs.JSON`` codec to
 encode object references defined as py:class:`~hdmf_zarr.utils.ZarrReference` dicts in datasets.
 Users may set the codec used to encode objects in Zarr datasets via the ``object_codec_class``
 parameter of the :py:func:`~hdmf_zarr.backend.ZarrIO.__init__` constructor of
--- a/src/hdmf_zarr/nwb.py
+++ b/src/hdmf_zarr/nwb.py
@@ -27,9 +27,10 @@ try:
                  'doc': 'a path to a namespace, a TypeMap, or a list consisting paths  to namespaces and TypeMaps',
                  'default': None})
         def __init__(self, **kwargs):
-            path, mode, manager, extensions, load_namespaces, synchronizer, storage_options = \
+            path, mode, manager, extensions, load_namespaces, synchronizer, storage_options, object_codec_class = \
                 popargs('path', 'mode', 'manager', 'extensions',
-                        'load_namespaces', 'synchronizer', 'storage_options', kwargs)
+                        'load_namespaces', 'synchronizer', 'storage_options',
+                        'object_codec_class', kwargs)
             if load_namespaces:
                 if manager is not None:
                     warn("loading namespaces from file - ignoring 'manager'")
@@ -53,7 +54,8 @@ try:
                                             manager=manager,
                                             mode=mode,
                                             synchronizer=synchronizer,
-                                            storage_options=storage_options)
+                                            storage_options=storage_options,
+                                            object_codec_class=object_codec_class)
 
         @docval({'name': 'src_io', 'type': HDMFIO, 'doc': 'the HDMFIO object for reading the data to export'},
                 {'name': 'nwbfile', 'type': 'NWBFile',
--- a/tests/unit/base_tests_zarrio.py
+++ b/tests/unit/base_tests_zarrio.py
@@ -17,7 +17,7 @@ from tests.unit.utils import (Baz, BazData, BazBucket, get_baz_buildmanager)
 
 # Try to import numcodecs and disable compression tests if it is not available
 try:
-    from numcodecs import Blosc, Delta, JSON
+    from numcodecs import Blosc, Delta, JSON, Pickle
     DISABLE_ZARR_COMPRESSION_TESTS = False
 except ImportError:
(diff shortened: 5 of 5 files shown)


src/hdmf_zarr/backend.py
```diff
@@ -94,7 +94,7 @@ def can_read(path):
              'default': None},
             {'name': 'object_codec_class', 'type': None,
              'doc': 'Set the numcodec object codec class to be used to encode objects.'
-                    'Use numcodecs.pickles.Pickle by default.',
```
oruebel, line 97: @magland just wondering if there may be some commits missing. I see that the comment in the docval string is changed but it looks like in line 124, ``self.__codec_cls = numcodecs.pickles.Pickle ...`` is still being set to Pickle in the PR.

oruebel (commented):
Could you also please add a line to the CHANGELOG https://github.com/hdmf-dev/hdmf-zarr/blob/dev/CHANGELOG.md

oruebel:
> In this PR I adjusted the default object_codec_class to numcodecs.JSON.

I think the main reason I didn't use JSON at the time was because of cases where objects are not JSON serializable. However, I don't remember whether that was an actual issue I encountered or whether I was just being overly cautious and used Pickle because I knew it would work. But I think if the test suite passes we should be Ok.

magland:
> Could you also please add a line to the CHANGELOG https://github.com/hdmf-dev/hdmf-zarr/blob/dev/CHANGELOG.md

Do you want me to add it under the 0.6.0 heading, or make a new heading? (0.6.1, 0.7.0)

oruebel:
> Do you want me to add it under the 0.6.0 heading, or make a new heading? (0.6.1, 0.7.0)

Please add it under a new heading "0.7.0 (upcoming)" since 0.6.0 has already been released

magland:
> > Do you want me to add it under the 0.6.0 heading, or make a new heading? (0.6.1, 0.7.0)
> 
> Please add it under a new heading "0.7.0 (upcoming)" since 0.6.0 has already been released

Done.

src/hdmf_zarr/backend.py
magland: I don't know how that happened. Strange. I added a commit.

magland:
Now that I made the actual modification, some tests are failing on my fork. So I guess the process is going to be more involved.

oruebel:
> Now that I made the actual modification, some tests are failing on my fork. So I guess the process is going to be more involved.

It looks like there are two main types of errors related to the JSON encoding:

1. `` TypeError: Object of type void is not JSON serializable``  --> This appears in 5 failing tests, which all seem to be related to writing ``ZarrWriteUnit``, i.e., I think these are likely all the same issue.  
2. `` AttributeError: 'int' object has no attribute 'append'`` --> This appears in 18 failing tests, and as far as I can tell they seem to be all related to writing compound dataset with references, more specifically, these I believe are writing a compound data type that is a mix of strings and int, similar to `TimeSeriesReferenceVectorData` in NWB.

oruebel:
> 2\. ` AttributeError: 'int' object has no attribute 'append'`

The point of failure for this error appears to be in all cases in the `write_dataset` function at:

https://github.com/hdmf-dev/hdmf-zarr/blob/556ed123e5cc07a434fd13593c06dca861267b8f/src/hdmf_zarr/backend.py#L1055-L1061

https://github.com/hdmf-dev/hdmf-zarr/pull/146 updated the writing of compound datasets to fix the determination of the dtype and may be relevant here.

magland:
Thanks, I'm working on a potential solution.

magland:
@oruebel I have pushed a solution where it resorts to a Pickle codec in certain (hopefully rare) cases, and prints a warning.

The cases are: (a) structured array and (b) compound dtype with embedded references

For the purposes of neurosift, I am okay with this, because it seems to be a rare occurrence. But I wasn't sure whether you wanted to either suppress the warnings or try to serialize this in a different way.

oruebel:
Thanks a lot! I think it would be great to see if we can use JSON for those cases too. The TimeSeriesReferenceVectorData is probably the main case in NWB. This is a compound of two ints and a reference. References are stored as a jSOn string, so I think those should be serializable with JSON too.

@mavaylon1 could you take a look at this PR to see if we can use JSON serialization for the compound data case that @magland mentioned as well. It would be good to be consistent in how we serialize objects. I think thus is related to the recent PR on compound data types.

magland:
I found this particularly challenging example in the wild. It seems it has a dataset that is a list of lists with embedded references.

000713 - Allen Institute - Visual Behavior - Neuropixels

https://neurosift.app/?p=/nwb&url=https://api.dandiarchive.org/api/assets/b2391922-c9a6-43f9-8b92-043be4015e56/download/&dandisetId=000713&dandisetVersion=draft

https://api.dandiarchive.org/api/assets/b2391922-c9a6-43f9-8b92-043be4015e56/download/


intervals/grating_presentations/timeseries

```
[[3, 135, <HDF5 object reference>], [138, 348, <HDF5 object reference>], [486, 392, <HDF5 object reference>], [878, 474, <HDF5 object reference>], [1352, 373, <HDF5 object reference>], [1725, 394, <HDF5 object reference>], [2119, 428, <HDF5 object reference>], [2547, 363, <HDF5 object reference>], [2910, 396, <HDF5 object reference>], [3306, 430, <HDF5 object reference>], [3736, 405, <HDF5 object reference>], [4141, 400, <HDF5 object reference>], [4541, 495, <HDF5 object reference>], [5036, 372, <HDF5 object reference>], [5408, 528, <HDF5 object reference>], [5936, 369, <HDF5 object reference>], [6305, 358, <HDF5 object reference>], [6663, 347, <HDF5 object reference>], [7010, 401, <HDF5 object reference>], [7411, 372, <HDF5 object reference>], [7783, 441, <HDF5 object reference>], [8224, 478, <HDF5 object reference>], [8702, 428, <HDF5 object reference>], [9130, 490, <HDF5 object reference>], [9620, 426, <HDF5 object reference>], [10046, 642, <HDF5 object reference>], [10688, 454, <HDF5 object reference>], [11142, 366, <HDF5 object reference>], [11508, 570, <HDF5 object reference>], [12078, 594, <HDF5 object reference>], [12672, 387, <HDF5 object reference>], [13059, 484, <HDF5 object reference>], [13543, 586, <HDF5 object reference>], [14129, 627, <HDF5 object reference>], [14756, 374, <HDF5 object reference>], [15130, 390, <HDF5 object reference>], [15520, 362, <HDF5 object reference>], [15882, 414, <HDF5 object reference>], [16296, 624, <HDF5 object reference>], [16920, 353, <HDF5 object reference>], [17273, 474, <HDF5 object reference>], [17747, 416, <HDF5 object reference>], [18163, 413, <HDF5 object reference>], [18576, 453, <HDF5 object reference>], [19029, 355, <HDF5 object reference>], [19384, 369, <HDF5 object reference>], [19753, 478, <HDF5 object reference>], [20231, 499, <HDF5 object reference>], [20730, 391, <HDF5 object reference>], [21121, 419, <HDF5 object reference>], [21540, 476, <HDF5 object reference>], [22016, 377, <HDF5 object reference>], [22393, 393, <HDF5 object reference>], [22786, 593, <HDF5 object reference>], [23379, 414, <HDF5 object reference>], [23793, 385, <HDF5 object reference>], [24178, 477, <HDF5 object reference>], [24655, 566, <HDF5 object reference>], [25221, 363, <HDF5 object reference>], [25584, 510, <HDF5 object reference>], [26094, 633, <HDF5 object reference>], [26727, 651, <HDF5 object reference>], [27378, 598, <HDF5 object reference>], [27976, 351, <HDF5 object reference>], [28327, 465, <HDF5 object reference>], [28792, 347, <HDF5 object reference>], [29139, 458, <HDF5 object reference>], [29597, 408, <HDF5 object reference>], [30005, 534, <HDF5 object reference>], [30539, 368, <HDF5 object reference>], [30907, 411, <HDF5 object reference>], [31318, 683, <HDF5 object reference>], [32001, 516, <HDF5 object reference>], [32517, 532, <HDF5 object reference>], [33049, 379, <HDF5 object reference>], [33428, 395, <HDF5 object reference>], [33823, 413, <HDF5 object reference>], [34236, 485, <HDF5 object reference>], [34721, 457, <HDF5 object reference>], [35178, 386, <HDF5 object reference>], [35564, 554, <HDF5 object reference>], [36118, 473, <HDF5 object reference>], [36591, 505, <HDF5 object reference>], [37096, 353, <HDF5 object reference>], [37449, 388, <HDF5 object reference>], [37837, 355, <HDF5 object reference>], [38192, 361, <HDF5 object reference>], [38553, 380, <HDF5 object reference>], [38933, 512, <HDF5 object reference>], [39445, 403, <HDF5 object reference>], [39848, 544, <HDF5 object reference>], [40392, 351, <HDF5 object reference>], [40743, 440, <HDF5 object reference>], [41183, 387, <HDF5 object reference>], [41570, 359, <HDF5 object reference>], [41929, 485, <HDF5 object reference>], [42414, 429, <HDF5 object reference>], [42843, 404, <HDF5 object reference>], [43247, 417, <HDF5 object reference>], [43664, 458, <HDF5 object reference>], [44122, 443, <HDF5 object reference>], [44565, 422, <HDF5 object reference>], [44987, 410, <HDF5 object reference>], [45397, 510, <HDF5 object reference>], [45907, 526, <HDF5 object reference>], [46433, 638, <HDF5 object reference>], [47071, 447, <HDF5 object reference>], [47518, 456, <HDF5 object reference>], [47974, 352, <HDF5 object reference>], [48326, 493, <HDF5 object reference>], [48819, 392, <HDF5 object reference>], [49211, 420, <HDF5 object reference>], [49631, 391, <HDF5 object reference>], [50022, 377, <HDF5 object reference>], [50399, 366, <HDF5 object reference>], [50765, 428, <HDF5 object reference>], [51193, 351, <HDF5 object reference>], [51544, 356, <HDF5 object reference>], [51900, 347, <HDF5 object reference>], [52247, 656, <HDF5 object reference>], [52903, 427, <HDF5 object reference>], [53330, 430, <HDF5 object reference>], [53760, 210, <HDF5 object reference>]]
```

oruebel:
> I found this particularly challenging example in the wild. It seems it has a dataset that is a list of lists with embedded references.

Yes, that is a `TimeSeriesReferenceVectorData` https://nwb-schema.readthedocs.io/en/stable/format.html#timeseriesreferencevectordata which is a compound dataset which has a reference to a `TimeSeries` and then two ints with the start and stop index.

oruebel:
> I found this particularly challenging example in the wild.

Note, in Zarr, the object reference is being represented as a JSON, so in Zarr the dtype for `TimeSeriesReferenceVectorData` ends up being a `(int, int, str)`. `hdmf-zarr` then deals with translating the  SON string part back to resolve the reference. This is done lazily on read. I.e., from a Zarr perspective, I think this datatype of `(int, int, str)` should in principle still be serializable in JSON.

src/hdmf_zarr/backend.py
```diff
@@ -1268,6 +1273,23 @@ def __list_fill__(self, parent, name, data, options=None):  # noqa: C901
...
+        # If we do, then we are going to resort to pickling the data and
+        # printing a warning.
+        has_structured_array = False
+        if dtype == object:
+            for c in np.ndindex(data_shape):
+                o = data
+                for i in c:
+                    o = o[i]
+                if isinstance(o, np.void) and o.dtype.names is not None:
+                    has_structured_array = True
```
sneakers-the-rat, line 1286: this seems pretty slow, iterating through every value in the array to check type. in structured arrays, wouldn't you just have to check the dtype of the whole array? in any case shouldn't we `break` the first time we get a `True`?

sneakers-the-rat: also couldn't this switch happen here: https://github.com/hdmf-dev/hdmf-zarr/blob/29adcca1af14bb03255fdf52723daedf6c67ce44/src/hdmf_zarr/backend.py#L1254-L1258 ?

sneakers-the-rat:
Can we turn the codec class into a getter method so we can encapsulate all switching logic in one place? 

currently there's: https://github.com/hdmf-dev/hdmf-zarr/blob/29adcca1af14bb03255fdf52723daedf6c67ce44/src/hdmf_zarr/backend.py#L154

the logic of when to use which codec is split in a few places, but it seems to ultimately depend on `builder.dtype`:

this check here: https://github.com/hdmf-dev/hdmf-zarr/blob/29adcca1af14bb03255fdf52723daedf6c67ce44/src/hdmf_zarr/backend.py#L1054
has its roots here: https://github.com/hdmf-dev/hdmf-zarr/blob/29adcca1af14bb03255fdf52723daedf6c67ce44/src/hdmf_zarr/backend.py#L955

and similarly this check: https://github.com/hdmf-dev/hdmf-zarr/blob/29adcca1af14bb03255fdf52723daedf6c67ce44/src/hdmf_zarr/backend.py#L1279
has its roots here: https://github.com/hdmf-dev/hdmf-zarr/blob/29adcca1af14bb03255fdf52723daedf6c67ce44/src/hdmf_zarr/backend.py#L1218
called from here: https://github.com/hdmf-dev/hdmf-zarr/blob/29adcca1af14bb03255fdf52723daedf6c67ce44/src/hdmf_zarr/backend.py#L1071

which is just the other leg of the above check, and L1276-1291 is very similar to L1037-1050.

`write_builder` and the other `write_*` methods are *almost* private methods since the docs only demonstrate using `write()`, but the builder is passed through them rather than stored as an instance attr (presumably since it gets remade on each `write()`? not really sure) so we can't assume we have it and use the above `object_codec_class` property, but it looks like every time we get `__codec_cls` we also have the builder or the `options` dict that gets constructed from it in scope, so couldn't we do something like `get_codec_cls(dtype)` ?

not my package, but when i was working on this before i had left this note: https://github.com/hdmf-dev/hdmf-zarr/blob/29adcca1af14bb03255fdf52723daedf6c67ce44/src/hdmf_zarr/backend.py#L1033-L1036 , which seems to be out of date bc we're using objects rather than converting to strings (presumably the root if the issue here), and so it also seems like we can consolidate the several places where we create and check a dtype into that `__resolve_dtype_helper__` method which could then make `get_codec_cls()` well defined as well.

oruebel:
@sneakers-the-rat some more details below, but to make a long story short, my hope is that we can:

1. Avoid having to use a mix of different codecs in the same file and as such avoid the need for logic to switch between codecs if possible
2. If we really need to use multiple codecs, then:
      * Yes, we should encapsulate the logic for selecting codes in a function, as you proposed
      * We should remove the `object_codec_class`   parameter on the `ZarrIO.__init__` method and not allow users to manually specify the codec to use (this is probably a good idea in general to avoid having to deal with to many corner cases)  

@magland @sneakers-the-rat sorry that I have not had the chance to dive into the code for this PR yet, but those are things I hope we can do before merging this PR, i.e., try to do **1.** if possible and implement **2.** if we can't make **1.** work.

> the logic of when to use which codec is split in a few places

TBH, my hope is that we can avoid having to use different codecs if possible. The use of `numcodecs.Pickle` was convenient, because it is pretty reliable in handling even complex cases, however, it makes it hard for non-Python codes (e.g., from JavaScript in the case of neurosift) to use the Zarr files directly. `numcodecs.JSON` makes that easier, but JSON is not as flexible as Pickle in terms of the cases it handles. The main issue seems to be structured arrays. However, in most cases those are a combination of ints and strings in our case, so my hope is that we can handle those in JSON as well, and only fall back to `Pickle` when absolutely necessary. 

> Can we turn the codec class into a getter method so we can encapsulate all switching logic in one place?

Currently the backend uses a single codec throughout, so a getter was not need, but if we really need to use multiple codes (which, again, I ideally would like to avoid if possible), then I totally agree that encapsulating this logic in a method is a good idea. We probably, then also need to get rid of `object_codec_class` as a parameter on `__init__` since we need  to control when to use what codec https://github.com/hdmf-dev/hdmf-zarr/blob/897d3b95ddfadd9364b2e2bc019e6ade86e920ce/src/hdmf_zarr/backend.py#L95-L97

> so couldn't we do something like `get_codec_cls(dtype)`

Yes, I believe that should work. The `object_codec_class` property is mainly there since the codec is currently configurable on  `__init__`.

mavaylon1:
@oruebel I'll dive in this week.

mavaylon1:
Recap and Updates:
We ideally only want one codec. I've been trying to get the compound case for references to work but I've ran into some issues. Using the dev branch of numcodecs, the problem is decoding the data. I made an issue ticket on the numcodec github and will try to find a fix. https://github.com/zarr-developers/numcodecs/issues/518

mavaylon1:
I have an "almost" proof of concept for this. I will share this week if it holds up with more exploration.

PR #183, opened 2024-04-18
Add sdupport for r- read mode to read without consolidated metadata

oruebel:
## Motivation

For benchmarking and when consolidated metadata may be bad, it will be useful to force reading without consolidated metadata. This is needed for the nwb-benchmarks. 

This PR:
-  adds a new mode `r-` to force that a file is opened in `r` mode without using consolidated metadata. 
- fixes a  bug where `ZarrIO.__open_file_consolidated` used properties of `ZarrIO` instead of the provided input parameters. This did not cause any errors so far, since the values matched but is still something that could have caused issues later on.
- adds unit tests for `-r` read mode
- fixes a minor bug in`ZarrStoreTestCase` which assumed in the `tearDown` that the output file was always being created, so clean-up would fail if a test-case did not create a file. 

## Checklist

- [X] Did you update CHANGELOG.md with your changes?

### DIFF
--- a/src/hdmf_zarr/backend.py
+++ b/src/hdmf_zarr/backend.py
@@ -88,7 +88,8 @@ class ZarrIO(HDMFIO):
              'doc': 'the path to the Zarr file or a supported Zarr store'},
             {'name': 'manager', 'type': BuildManager, 'doc': 'the BuildManager to use for I/O', 'default': None},
             {'name': 'mode', 'type': str,
-             'doc': 'the mode to open the Zarr file with, one of ("w", "r", "r+", "a", "w-")'},
+             'doc': 'the mode to open the Zarr file with, one of ("w", "r", "r+", "a", "w-") '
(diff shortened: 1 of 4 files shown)


oruebel:
@mavaylon1 this is ready for review.

src/hdmf_zarr/backend.py
```diff
@@ -161,7 +162,7 @@ def open(self):
             # As a result, when in other modes, the file will not use consolidated metadata.
             if self.__mode not in ['r', 'r+']:
                 self.__file = zarr.open(store=self.path,
-                                        mode=self.__mode,
+                                        mode=self.__mode if self.__mode != 'r-' else 'r',
```
mavaylon1: @oruebel I think it is cleaner to not have this conditional statement in a function call. Could you move it out?

oruebel: Fixed

mavaylon1 (approved):

PR #184, opened 2024-04-26
Fix bug in ZarrIO.__open_consolidated 

oruebel:
## Motivation

`ZarrIO.__open_consolidated`` currently used the following check `if os.path.isfile(path+'/.zmetadata'):` to determine whether the file should be open with consolidated metadata or without. This check, however, assumes that the file is local and will not work for remote files stored on S3. 

## Changes

To address this issue this PR changes to a try/except logic instead. `zarr.open_consolidated` will raise a  `KeyError` if the `.zmetata` is missing so we can just try to open the file this way first and the fall back to the regular method if it fails. 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@
 ### Docs
 * Removed `linkable` from the documentation to keep in line with `hdmf-schema-language`. @mavaylon1 [#180](https://github.com/hdmf-dev/hdmf-zarr/pull/180)
 
-### Internal Fixes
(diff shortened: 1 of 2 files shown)


oruebel:
@mavaylon1 could you please review this PR as well for the next release.

Note, I unfortunately could not add a unit test for this issue, but I filed https://github.com/hdmf-dev/hdmf-zarr/issues/185 to address this gap in our tests separately .

mavaylon1 (approved):

PR #187, opened 2024-04-29
Update pyproject.toml

mavaylon1:
## Motivation

Numpy has released numpy 2.0, which came with deprecations. I'm going to pin below 2.0 for the release on April 30th. 

Update:
Numpy 2.0 supports python 3.9-3.12. I think we should keep this below 2.0 until October when python 3.8 is deprecated. We will fix the deprecations then, rather than make a change now and make another change in October.

## How to test the behavior?
```
```

## Checklist


mavaylon1:
Review Notes:
I ran the workflow for "run all tests" and it passes everything.
I did not update the changelog this is something we do not put there.

rly:
> Numpy 2.0 supports python 3.9-3.12. I think we should keep this below 2.0 until October when python 3.8 is deprecated. We will fix the deprecations then, rather than make a change now and make another change in October.

Many packages have dropped Python 3.8 support in their latest version. An alternative approach to waiting until October to support Numpy 2.0 is to update the code to work with both Numpy < 2.0 and Numpy >= 2.0 (I don't think they would conflict), allow both in the requirements, and set up `requirements.txt` to use a version of Numpy < 2.0 for Python 3.8 and the latest Numpy >= 2.0 version for other versions. Like in https://github.com/hdmf-dev/hdmf/blame/95ba8765af25b0ffce0f20c28b047eb1616873a5/requirements.txt

mavaylon1:
> > Numpy 2.0 supports python 3.9-3.12. I think we should keep this below 2.0 until October when python 3.8 is deprecated. We will fix the deprecations then, rather than make a change now and make another change in October.
> 
> Many packages have dropped Python 3.8 support in their latest version. An alternative approach to waiting until October to support Numpy 2.0 is to update the code to work with both Numpy < 2.0 and Numpy >= 2.0 (I don't think they would conflict), allow both in the requirements, and set up `requirements.txt` to use a version of Numpy < 2.0 for Python 3.8 and the latest Numpy >= 2.0 version for other versions. Like in https://github.com/hdmf-dev/hdmf/blame/95ba8765af25b0ffce0f20c28b047eb1616873a5/requirements.txt

Third option is to (for the releases tomorrow) just do this. Then do the release, then try to see if a middle ground for numpy 2.0 and below.

oruebel:
Changing the requirements to have different Numpy versions for different Python versions seems simple enough. I assume this would be just something like:

```
numpy>=1.24, <2.0';python_version<='3.8'   # pin below 2.0 until python 3.8 is deprecated
numpy>=1.24;python_version>'3.8'
```

mavaylon1:
> Changing the requirements to have different Numpy versions for different Python versions seems simple enough. I assume this would be just something like:
> 
> 
> 
> ```
> 
> numpy>=1.24, <2.0';python_version<='3.8'   # pin below 2.0 until python 3.8 is deprecated
> 
> numpy>=1.24;python_version>'3.8'
> 
> ```
> 
> 

So no matter what python version, we will get the error from numpy because np.string_ is called in the objectmapper but that is now deprecated. 

I still standby pinning below 2.0 for the release, then finding all the areas we need fixed to support numpy 2.0.

pyproject.toml
```diff
@@ -31,7 +31,7 @@ classifiers = [
 dependencies = [
     'hdmf>=3.9.0',
     'zarr>=2.11.0',
-    'numpy>=1.24',
+    'numpy>=1.24, <2.0', # pin below 2.0 until python 3.8 is deprecated
```
rly:
```suggestion
    'numpy>=1.24, <2.0', # pin below 2.0 until HDMF supports numpy 2.0
```

rly (approved):

PR #188, opened 2024-04-29
Update coverage call, don't install in editable mode in testing

rly:
## Motivation

Analogous to https://github.com/hdmf-dev/hdmf/pull/1107

https://setuptools.pypa.io/en/latest/userguide/development_mode.html says
> Editable installs are not a perfect replacement for regular installs in a test environment. When in doubt, please test your projects as installed via a regular wheel. There are tools in the Python ecosystem, like [tox](https://pypi.org/project/tox) or [nox](https://pypi.org/project/nox), that can help you with that (when used with appropriate configuration).

We use editable installs in testing in a couple places, mostly to help `coverage` find the sources, but we can configure that in `pyproject.toml` instead.

## How to test the behavior?
Run all GitHub actions

## Checklist


.github/workflows/run_coverage.yml
```diff
@@ -48,20 +48,26 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install -r requirements-dev.txt -r requirements.txt -r requirements-opt.txt
+          python -m pip install -r requirements-dev.txt -r requirements.txt
+
+      - name: Install optional dependencies
+        if: ${{ matrix.opt_req }}
+        run: python -m pip install -r requirements-opt.txt
```
rly, line 55: The testing matrix was set up to test and report coverage with both optional requirements installed and not installed. Testing with the optional requirements takes a lot longer because of the streaming test.

mavaylon1 (approved):

Issue #179, opened 2024-03-27
[Bug]: `export` fails to correctly save units after adding columns


alejoe91:
### What happened?

When using the `io.export` after appending some unit columns, the resulting NWB-zarr file only has the newly added columns and it's missing the original ones (including `spike_times`). This makes the exported NWB invalid.

### Steps to Reproduce

```shell
with NWBZarrIO(str(nwb_input_path), "r") as read_io:
    nwbfile = read_io.read()

    for col in some_df_with_data.columns:
        nwbfile.add_unit_column(
            name=col,
            description=col,
            data=some_df_with_data[col].values
        )

    with NWBZarrIO(str(nwb_output_path), "w") as export_io:
        export_io.export(src_io=read_io, nwbfile=nwbfile)

# loading the exported NWB zarr folder fails
io = NWBZarrIO(str(nwb_output_path), "r")
nwbfile = io.read()
```

### Traceback

```shell
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
File /opt/conda/lib/python3.9/site-packages/hdmf_zarr/backend.py:729, in ZarrIO.resolve_ref(self, zarr_ref)
    728 try:
--> 729     target_zarr_obj = target_zarr_obj[object_path]
    730 except Exception:

File /opt/conda/lib/python3.9/site-packages/zarr/hierarchy.py:500, in Group.__getitem__(self, item)
    499 else:
--> 500     raise KeyError(item)

KeyError: '/units/spike_times'

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
Cell In[71], line 2
      1 io = NWBZarrIO(str(nwb_output_path), "r")
----> 2 nwbfile = io.read()

File /opt/conda/lib/python3.9/site-packages/hdmf/utils.py:668, in docval.<locals>.dec.<locals>.func_call(*args, **kwargs)
    666 def func_call(*args, **kwargs):
    667     pargs = _check_args(args, kwargs)
--> 668     return func(args[0], **pargs)

File /opt/conda/lib/python3.9/site-packages/hdmf/backends/io.py:56, in HDMFIO.read(self, **kwargs)
     53 @docval(returns='the Container object that was read in', rtype=Container)
     54 def read(self, **kwargs):
     55     """Read a container from the IO source."""
---> 56     f_builder = self.read_builder()
     57     if all(len(v) == 0 for v in f_builder.values()):
[...]

PR #194, opened 2024-05-16
Broken Links when Exporting

mavaylon1:
## Motivation

This is a refactor of export to formally define mechanisms for references and internal/external links. This Pr is the first stage as it covers only Zarr to Zarr export and does not yet support compound cases. 

This also removes TempStore from the export test suite in favor of files that will be created and removed during testing. This helps with debugging and also allows for behavior that better matches the real world.

Export Cases go as follows:

1. Internal References/Links need to be preserved and not made into ExternalReferences
2. External Links need to be preserved to the correct target.
3. We do not support External References.

## How to test the behavior?
```
```

## Checklist


mavaylon1:
More Notes:
FileA is a HDF5 file and is exported to FileB, a zarr file.
FileA has both internal links, external links, and references (which are always internal for us). I remember you saying we don't do external links, but maybe my memory is off. I ask because the export doc talks about external links.
Because the backends are different, everything is copied over. Does that mean during export, FileB will have copies of what is being externally linked inside FileB? That also means every internal link and every reference is also preserved.
Now I export FileB to FileC, which is still zarr. However, I add a few containers and append to existing containers. What is happening right now in zarr is that new data is added correctly; however, we now create a link to the FileB from FileC if we don't specify copy. This is wrong in that this isn't what export is supposed to do as external links can break easily from moving files. (which is why I ask about external links and why is it talked about in the tutorial). What we need to do is copy everything to FileC, preserving internal links and references.


If FileA contains an external link to a dataset in FileX, then FileB should also contain an external link to the dataset in FileX.
[6:23](https://nwb-users.slack.com/archives/D01K4SQHZ1N/p1719624238083329)
Same for FileB and FileC.
6:24
If FileB is read, a new external link is added, and then the file is exported to FileC, then it is written as an external link
6:25
If FileB is read, a new external link is added, and then the file is exported to FileC with write_args={'link_data': False}, then the linked dataset is copied
[6:25](https://nwb-users.slack.com/archives/D01K4SQHZ1N/p1719624313155999)
This is a very specific, niche case

mavaylon1:
Goal for this PR:
1. We want to make sure the external links are just copied when we export, not preserved as links. 
- When export Zarr File A to Zarr File B and add stuff to existing containers, they should all not be links. (this is what Alessio needs)

oruebel:
Just in case this is relevant for this PR. The following test cases mirror tests from HDMF but were disabled in the hdfm_zarr test suite because links on export didn't fully work. If this PR fixes this, then we should also look at updating these tests.

https://github.com/hdmf-dev/hdmf-zarr/blob/8ca578733db5078d1ff6d2dfb47407a680c58caf/tests/unit/base_tests_zarrio.py#L1329-L1521

https://github.com/hdmf-dev/hdmf-zarr/blob/8ca578733db5078d1ff6d2dfb47407a680c58caf/tests/unit/test_io_convert.py#L993-L1069

mavaylon1:
> Just in case this is relevant for this PR. The following test cases mirror tests from HDMF but were disabled in the hdfm_zarr test suite because links on export didn't fully work. If this PR fixes this, then we should also look at updating these tests.
> 
> https://github.com/hdmf-dev/hdmf-zarr/blob/8ca578733db5078d1ff6d2dfb47407a680c58caf/tests/unit/base_tests_zarrio.py#L1329-L1521
> 
> https://github.com/hdmf-dev/hdmf-zarr/blob/8ca578733db5078d1ff6d2dfb47407a680c58caf/tests/unit/test_io_convert.py#L993-L1069

Good to know. I believe my tests are similar if not the same ones. Thanks for pointing this out so we don't have duplicates.

mavaylon1:
Related Issues: #179 #205

mavaylon1:
Review Notes:
This is the first stage PR. Some tests are still commented out, but the changes should be enough to unblock the Allen Institute as long as they do not have compound cases (TBD) and backend changes (they do not).

Things are subject to change during a refactor that will be done before the end of November. This is to really checkpoint the work and have the team be on the same page with the changes. That being said, optimizations will be made and the final product is what will lead to a large release of hdmf-zarr. 

I don't have an issue ticket this directly fixes. However, we do have some that are related. #179

tests/unit/base_tests_zarrio.py
```diff
@@ -358,10 +357,6 @@ def test_write_references_roundtrip(self):
                 # Check the ZarrReference first
                 self.assertEqual(zarr_ref.object_id, expected_value['object_id'])
                 self.assertEqual(zarr_ref.source_object_id, expected_value['source_object_id'])
-                # Check that the ZarReference is being resolved via the ZarrIO.resolve_ref
```
mavaylon1, line 364: not needed

tests/unit/base_tests_zarrio.py
```diff
@@ -1120,108 +1110,103 @@ def test_soft_link_group(self):
...
             read_foofile2 = read_io.read()
-            # make sure the linked group is within the same file
-            self.assertEqual(read_foofile2.foo_link.container_source, self.store_path[1])
-            zarr_linkspec1 = zarr.open(self.store_path[0])['links'].attrs.asdict()['zarr_link'][0]
-            zarr_linkspec2 = zarr.open(self.store_path[1])['links'].attrs.asdict()['zarr_link'][0]
-            self.assertEqual(zarr_linkspec1.pop('source'), ".")
-            self.assertEqual(zarr_linkspec2.pop('source'), ".")
-            self.assertDictEqual(zarr_linkspec1, zarr_linkspec2)
-
-    def test_soft_link_dataset(self):
```
mavaylon1, line 1143: This was refactored in a different test.

tests/unit/base_tests_zarrio.py
```diff
@@ -1601,6 +1571,5 @@ def test_wrong_mode(self):
 
         with ZarrIO(self.store[0], mode='r') as read_io:
             with ZarrIO(self.store[1], mode='a') as export_io:
-                msg = "Cannot export to file %s in mode 'a'. Please use mode 'w'." % self.store_path[1]
-                with self.assertRaisesWith(UnsupportedOperation, msg):
+                with self.assertRaises(UnsupportedOperation):
```
mavaylon1, line 1597: This is a cheeky change I know, but the msg is the same with some weird formatting that I don't think time needs to be spent fixing.

tests/unit/test_io_convert.py
```diff
@@ -53,10 +53,9 @@
 
 from tests.unit.utils import (Foo, FooBucket, FooFile, get_foo_buildmanager,
                               Baz, BazData, BazBucket, get_baz_buildmanager,
-                              BazCpdData, get_temp_filepath)
```
mavaylon1, line 56: This is because CPD cases are the next stage

tests/unit/test_zarrio.py
```diff
@@ -86,32 +89,6 @@ def setUp(self):
         self.store = [DirectoryStore(p) for p in self.store_path]
 
 
-#########################################
```
mavaylon1, line 90: Removed TempStore because too much time was going into getting it to work. Not something users even use.

tests/unit/test_zarrio.py
```diff
@@ -272,3 +249,152 @@ def test_append_references(self):
             read_container = append_io.read()
             self.assertEqual(len(read_container.baz_data.data), 11)
             self.assertIs(read_container.baz_data.data[10], read_container.bazs["new"])
+
+
+class TestExport(TestCase):
```
mavaylon1: I added this here, but it could be moved to the base test file. I'm fine with either.

tests/unit/base_tests_zarrio.py
mavaylon1: test_soft_links_export

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # HDMF-ZARR Changelog
 
+## 1.0.0 (Upcoming)
+### Enhancements
+# Add initial refactor of export, supporting references and internal/external links from Zarr to Zarr. @mavaylon1 [#194](https://github.com/hdmf-dev/hdmf-zarr/pull/194)
```
```suggestion
* Added initial refactor of export, supporting references and internal/external links from Zarr to Zarr. @mavaylon1 [#194](https://github.com/hdmf-dev/hdmf-zarr/pull/194)
```

src/hdmf_zarr/backend.py
```diff
@@ -546,14 +547,41 @@ def write_group(self, **kwargs):
                     export_source=export_source,
                 )
 
-        # write all links (haven implemented)
         links = builder.links
         if links:
+            # Note: In HDMF, we simply call write_link (within write_group) which contains a similar logic as seen
+            # below. This is not the case here (until a refactor).
```
oruebel: It looks like the only place where `write_link` is being called is in this for loop. Why not move the logic to `write_link` directly now? I think that would help keep the code much more readable as it keeps the link logic in one place.

src/hdmf_zarr/backend.py
```diff
@@ -546,14 +547,41 @@ def write_group(self, **kwargs):
...
+                        # Case 1:
+                        # sub_builder.builder.source == export_source
+                        # This means we have a SoftLink for a group and so we want the exported link to
+                        # also point "inwards" in the file being created.
+                        #################################
+                        # Case 2:
+                        # sub_builder.builder.source == group_filename
+                        # This is still a SoftLink; however, it is from adding a link to a group after FileA
+                        # has been read and we are exporting that to FileB. We still want the link to be "inwards".
+                        source = group_filename
```
oruebel: It's confusing that the variable is called `source` but then it is used to set `export_source` on the `write_link` function. Moving the logic to `write_link` will probably help, be we should name the variable according to what it is used for, since this is not the actual "source" you are using.

src/hdmf_zarr/backend.py
```diff
@@ -718,7 +754,7 @@ def resolve_ref(self, zarr_ref):
             source_file = str(zarr_ref['source'])
         # Resolve the path relative to the current file
         if not self.is_remote():
-            source_file = os.path.abspath(os.path.join(self.source, source_file))
+            source_file = os.path.abspath(source_file)
```
oruebel, line 753: Can you clarify why `self.source` does not need to appear here? Isn't `source_file` a path relative to this file (i.e., `self.source`)?

tests/unit/test_zarrio.py
```diff
@@ -86,32 +89,6 @@ def setUp(self):
...
-        self.store = TempStore()
-        self.store_path = self.store.path
-
-
-class TestExportZarrToZarrTempStore(BaseTestExportZarrToZarr):
-    """Test exporting Zarr to Zarr using TempStore."""
-    def setUp(self):
-        super().setUp()
-        self.store = [TempStore() for i in range(len(self.store_path))]
-        self.store_path = [s.path for s in self.store]
```
oruebel, line 113: I think it's Ok to remove TempStore. Maybe we declare it as deprecated and raise a DeprecationWarning in ZarrIO.__init__ , so that we can remove it with the next main release.

oruebel:
I'm still making my way though the changes in `backend.py`. Just to double-check, are the changes here compatible with existing NWB Zarr files? As far as I can tell, the answer is probably yes, but wanted to check. 

You may also want to check that the docs at https://github.com/hdmf-dev/hdmf-zarr/blob/dev/docs/source/storage.rst don't need to be updated for this.

mavaylon1:
> I'm still making my way though the changes in `backend.py`. Just to double-check, are the changes here compatible with existing NWB Zarr files? As far as I can tell, the answer is probably yes, but wanted to check.
> 
> You may also want to check that the docs at https://github.com/hdmf-dev/hdmf-zarr/blob/dev/docs/source/storage.rst don't need to be updated for this.

So I believe any file with just internal links and references should be fine. Anything with External may not work.

oruebel:
> So I believe any file with just internal links and references should be fine. Anything with External may not work.

In that case, I would ping Alessio to see if he can test out the PR to make sure it works with their files

mavaylon1:
> > So I believe any file with just internal links and references should be fine. Anything with External may not work.
> 
> In that case, I would ping Alessio to see if he can test out the PR to make sure it works with their files

I will ping him once we are solid on this PR. Let me address your comments and we can shoot to ping him Monday!

src/hdmf_zarr/backend.py
mavaylon1: Hmmm I see what you are saying.

src/hdmf_zarr/backend.py
mavaylon1: In HDMFIO self.source is set as an absolute path
```
manager, source, herd_path = getargs('manager', 'source', 'herd_path', kwargs)
        if isinstance(source, Path):
            source = source.resolve()
        elif (isinstance(source, str) and
              not (source.lower().startswith("http://") or
                   source.lower().startswith("https://") or
                   source.lower().startswith("s3://"))):
            source = os.path.abspath(source)

        self.__manager = manager
        self.__built = dict()
        self.__source = source
```
source_file is indeed just a relative path. 

Having them joined together leads to something like this : `'/Users/mavaylon/Research/NWB/hdmf-zarr/test_io.zarr/test_io.zarr'
`

Now what if you want to use the directory of self.source? so say something like os.path.abspath(os.path.join(os.dirname(self.source), source_file))? This assumes the self.source and the source_file are in the same directory. They don't have to be. source_file is the the source of the zarr reference. self.source is just what we are exporting from.

src/hdmf_zarr/backend.py
mavaylon1: I think what I should do is actually have this in _create_ref and modify write_link slightly to take in the ref output from create_ref. We call create_ref in multiple places, including write_link.

mavaylon1: This will clean up the logic here but also from 623-627

src/hdmf_zarr/backend.py
mavaylon1: I did a name change. I will need to to another round of checks to make sure I carried it through, but I think this should help distinguish the variables more.

src/hdmf_zarr/backend.py
```diff
@@ -599,17 +599,11 @@ def write_attributes(self, **kwargs):
                         raise TypeError(str(e) + " type=" + str(type(value)) + "  data=" + str(value)) from e
             # Case 2: References
             elif isinstance(value, (Container, Builder, ReferenceBuilder)):
-                # TODO: Region References are not yet supported
-                # if isinstance(value, RegionBuilder):
-                #     type_str = 'region'
-                #     refs = self._create_ref(value.builder)
                 if isinstance(value, (ReferenceBuilder, Container, Builder)):
```
rly: collapse the if statements and see why container is there

src/hdmf_zarr/backend.py
```diff
@@ -671,6 +665,9 @@ def get_zarr_paths(zarr_object):
         # path from the filepath to the object
         objectpath = "/" + os.path.relpath(fullpath, filepath)
         # return the result
+        if relative:
```
rly: Revert change and move the function to the change.

src/hdmf_zarr/backend.py
```diff
@@ -796,22 +786,36 @@ def _create_ref(self, ref_object, export_source=None):
...
+            # be either a (str, Path, *SUPPORTED_ZARR_STORES). That being said,
+            # when self.path is a Path, it is converted to a str in __init__.
+            # We only have to deal with *SUPPORTED_ZARR_STORES and strings.
+            ref_link_source = ref_link_source.path
+
+        # Note: We want want to construct the relative path with
+        # os.path.relpath(<absolute_path_to_the_target>, <absolute_path_to_the_file_that_is_being_exported_to>)
+        # That being said, we want to avoid a reference being defined as '.' because '.' means whatever file you
+        # are in. This does not help if you are trying to access a link/ref in another file and the source says
+        # '.' so look in yourself. That is why the dirname is there.
```
rly, line 832: Document this in two places: Export doc and Major Release 1.0

src/hdmf_zarr/backend.py
mavaylon1: I reverted the change, but I like it as a static method.

src/hdmf_zarr/backend.py
```diff
@@ -789,24 +784,22 @@ def _create_ref(self, ref_object, export_source=None):
...
         if isinstance(ref_object, Builder):
             if isinstance(ref_object, LinkBuilder):
-                builder = ref_object.target_builder
+                builder = ref_object.builder
             else:
                 builder = ref_object
         elif isinstance(ref_object, ReferenceBuilder):
             builder = ref_object.builder
-        else:
-            builder = self.manager.build(ref_object)
```
rly, line 798: If `ref_object` is a `Container`, which it seems like it can be based on line 635, then I think we should keep these lines. If it is not possible for ref_object to be a `Container`, then please update the if statement on line 635.

mavaylon1: good catch

src/hdmf_zarr/backend.py
```diff
@@ -1047,6 +1070,8 @@ def write_dataset(self, **kwargs):  # noqa: C901
             refs = list()
             type_str = list()
             for i, dts in enumerate(options['dtype']):
+                # TODO: This appears to be code to distinguish between region and object References
+                # and will be refactored/removed.
```
```suggestion
```

rly: The referenced RegionReference code was removed in https://github.com/hdmf-dev/hdmf-zarr/pull/236

tests/unit/base_tests_zarrio.py
```diff
@@ -1123,108 +1113,102 @@ def test_soft_link_group(self):
...
-                export_io.export(src_io=read_io, write_args=dict(link_data=False))
-        print(zarr.open(self.store_paths[0]).tree())
-        print(zarr.open(self.store_paths[1]).tree())
-        with ZarrIO(self.store_paths[1], manager=get_foo_buildmanager(), mode='r') as read_io:
-            read_foofile2 = read_io.read()
-            # make sure the linked dataset is within the same file
-            print(open(self.source_paths[1]+"/buckets/bucket1/foo_holder/foo1/.zattrs", 'r').read())
-            self.assertEqual(read_foofile2.foofile_data.path, self.source_paths[1])
-        """
+            if isinstance(self.store[1], str):
```
rly: When would `self.store[1]` not be a string? In setup(), it is `self.store = [f"file{i}.zarr" for i in range(3)]`

tests/unit/base_tests_zarrio.py
```diff
@@ -1123,108 +1113,102 @@ def test_soft_link_group(self):
...
-        #print()
-        print(zarr.open(self.store_paths[1])['links'].attrs.asdict())
-        print(zarr.open(self.store_paths[2])['links'].attrs.asdict())
-        with ZarrIO(self.store_paths[2], manager=get_foo_buildmanager(), mode='r') as read_io:
+        with ZarrIO(self.store[2], manager=get_foo_buildmanager(), mode='r') as read_io:
             read_foofile2 = read_io.read()
             # make sure the linked group is read from the first file
-            self.assertEqual(read_foofile2.foo_link.container_source, self.source_paths[0])
-        """
+            if isinstance(self.store[0], str):
```
rly: See comment above -- when is `self.store[0]` ever not a string?

tests/unit/base_tests_zarrio.py
```diff
@@ -988,12 +983,7 @@ class BaseTestExportZarrToZarr(BaseZarrWriterTestCase):
     """
 
     def setUp(self):
-        self.store = [
-            get_temp_filepath(),
-            get_temp_filepath(),
-            get_temp_filepath(),
-            get_temp_filepath()
-        ]
+        self.store = [f"file{i}.zarr" for i in range(3)]
```
rly: What is the difference between `self.store` and `self.store_path`? Why have two different variables that are the same?

rly:
I made some minor comments regarding the tests. Otherwise looks good to me.

tests/unit/base_tests_zarrio.py
mavaylon1:
```
(Pdb) self.store
['file0.zarr', 'file1.zarr', 'file2.zarr']
(Pdb) n
--Return--
> /Users/mavaylon/Research/NWB/hdmf-zarr/tests/unit/base_tests_zarrio.py(988)setUp()->None
-> self.store_path = self.store
(Pdb) n
> /Users/mavaylon/Research/NWB/hdmf-zarr/tests/unit/test_zarrio.py(87)setUp()
-> self.store = [DirectoryStore(p) for p in self.store_path]
(Pdb) 
```
This is the setUP of test_zarrio.py. It covers the case when the store is a DirectoryStore since we support that it doesn't need to be just a str path.

rly: Oh, I see. It's overridden elsewhere. That's a little confusing TBH, but I can suggest an improvement in a separate PR.

rly (approved):

Issue #192, opened 2024-05-08
[Bug]: Zarr 2.18.0 with Blosc


CodyCBakerPhD:
### What happened?

Just encountered test failure in NeuroConv due to latest Zarr release on May 7

Full log: https://github.com/catalystneuro/neuroconv/actions/runs/9005172770/job/24739878400

Including test case below, show have most of what you need to reproduce

Wanted to check if this has anything to do with how the file is being read on hdmf-zarr side, or otherwise just let y'all be aware of the issue

### Steps to Reproduce

```shell
tmpdir = local('/tmp/pytest-of-runner/pytest-0/popen-gw1/test_simple_time_series_zarr_g0')
integer_array = array([[   606,  22977,  27598, ...,  21453,  14831,  29962],
       [-26530,  -9155,  -6666, ...,  18490,  -6943,   1...5954, -21319, ...,  -8983, -30074, -24446],
       [-30841, -12815,  28599, ...,  24069, -15762,  -3284]], dtype=int16)
case_name = 'generic'
iterator = <class 'neuroconv.tools.hdmf.SliceableDataChunkIterator'>
iterator_options = {}, backend = 'zarr'

    @pytest.mark.parametrize(
        "case_name,iterator,iterator_options",
        [
            ("unwrapped", lambda x: x, dict()),
            ("generic", SliceableDataChunkIterator, dict()),
            ("classic", DataChunkIterator, dict(iter_axis=1, buffer_size=30_000 * 5)),
            # Need to hardcode buffer size in classic case or else it takes forever...
        ],
    )
    @pytest.mark.parametrize("backend", ["hdf5", "zarr"])
    def test_simple_time_series(
        tmpdir: Path,
        integer_array: np.ndarray,
        case_name: str,
        iterator: callable,
        iterator_options: dict,
        backend: Literal["hdf5", "zarr"],
    ):
        data = iterator(integer_array, **iterator_options)
    
        nwbfile = mock_NWBFile()
        time_series = mock_TimeSeries(name="TestTimeSeries", data=data)
        nwbfile.add_acquisition(time_series)
    
        backend_configuration = get_default_backend_configuration(nwbfile=nwbfile, backend=backend)
        dataset_configuration = backend_configuration.dataset_configurations["acquisition/TestTimeSeries/data"]
        configure_backend(nwbfile=nwbfile, backend_configuration=backend_configuration)
    
[...]

PR #195, opened 2024-05-16
Update for zarr 2.18.0 and zarr 2.18.1

mavaylon1:
## Motivation

This PR is to update hdmf-zarr for changes that came n 2.18.0 and new changes in 2.18.0

2.18.0
This is in response to the failing workflows using the latest version of zarr, specifically the failing test test_fsspec_streaming.

The issue is how we are reading zarr scalar arrays. We are currently trying to directly index and access the scalar. From looking at the documentation, zarr accessed a scalar array as `z[:]`, which seems to solve the problem. I also saw that `z[()]` was another syntax that works. 

Why?
Not completely clear.
When we access a scalar array in numpy we do so with notation `[()]`. Supposedly, Zarr used to support indexing scalar arrays as `z[0]`, but that was updated to numpy standard earlier in one of the zarr version 2.#. Maybe this was a functionality that finally got flushed out fully. 

**2.18.1**
To ensure that the data being assigned is in a format that the Zarr array can handle efficiently, Zarr arrays seem to require that we set with numpy arrays, i.e., `dset[:] = np.array(data)`.

Another weird behavior is when setting a dataset of references, the current method puts the entire dataset in each index vs matching the individual reference dictionary to the index (which is what we expect). 

The solution to both is make sure the data is in an array first. 

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.7.0 (May 2, 2024)
 ### Bug Fixes
-* Fixed bug on how we access scalar arrays. Added warning filter for Zarr deprecation of NestedDirectoryStore. @mavaylon1 [#195](https://github.com/hdmf-dev/hdmf-zarr/pull/195)
+* Fixed bug on how we access scalar arrays. Added warning filter for Zarr deprecation of NestedDirectoryStore. Fixed bug on how we write a dataset of references. @mavaylon1 [#195](https://github.com/hdmf-dev/hdmf-zarr/pull/195)
 
 ## 0.7.0 (May 2, 2024)
 ### Enhancements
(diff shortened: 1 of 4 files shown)


mavaylon1:
Hey @oruebel can you give me your thoughts. I believe this is related to #192

oruebel:
Using the `[()]` syntax to access scalars makes sense to me

mavaylon1:
@oruebel There are also some deprecation warnings for zarr 3.0. I will add filters in this PR, but I'll also make a ticket for those deprecations to be resolved to whatever zarr will be pivoting to.

mavaylon1:
Review Notes:
I need to ping to see if this now resolved with this update or to see what changes:
https://github.com/hdmf-dev/hdmf-zarr/issues/192

rly (approved):
LGTM. Thanks for taking care of this.

PR #203, opened 2024-07-03
Append References

mavaylon1:
## Motivation

This is in conjunction with https://github.com/hdmf-dev/hdmf/pull/1135

We want to support appending a dataset of references. 

## How to test the behavior?
```
```

## Checklist


mavaylon1:
@rly take a look. We need to merge the PR for zarr optional first, then the HDMF ticket related to this, then this. If you also agree with my approach to the paths (Even though it is a little hacky, I can't think of a better way given our conversation over slack), then I can wrap this up with tests tomorrow.

mavaylon1:
Review Notes:

The ruff ignores are because changing it to "is" returns errors, i.e., the functionality changes for whatever reason. We can update this if we want, but not in this PR and it would be the lowest of priorities.

src/hdmf_zarr/zarr_utils.py
```diff
@@ -73,6 +74,18 @@ def __iter__(self):
...
+    def append(self, arg):
+        # Create Builder
+        builder = self.invert().io.manager.build(arg)
+
+        # Get File/Source Object ID
+        f_builder = self.invert().io.read_builder()
+        source_file = self.invert().io.manager.construct(f_builder)
+
+        # Create ZarrReference
+        ref = self.invert().io._ZarrIO__get_ref(builder, source_file=source_file)
```
rly: Is there a difference between `self.invert().io` and `self.io`? I'm trying to follow the logic and am getting lost.

mavaylon1: Yes. Did not catch that.

Logic:
I have a file that has a dataset of references. To add a new reference, I need to create one for my arg, which is a Container object. 

To create a references I use get_ref, which needs the builder for the arg.

tests/unit/test_zarrio.py
```diff
@@ -182,3 +185,47 @@ def test_force_open_without_consolidated_fails(self):
...
+            read_container = append_io.read()
+            DoR = read_container.baz_data.data
+            DoR.append(DoR[0])
+
+            expected =  {'source': '.', 'path': '/bazs/baz0',
+                         'object_id': DoR[0].object_id,
+                         'source_object_id': container.object_id}
+
+            self.assertEqual(len(read_container.baz_data.data), 11)
+            self.assertDictEqual(read_container.baz_data.data.dataset[10], expected)
```
rly: Just in case, could you test closing this file, opening it, and making sure the size is still 11?

src/hdmf_zarr/zarr_utils.py
```diff
@@ -73,6 +74,14 @@ def __iter__(self):
     def __next__(self):
         return self._get_ref(super().__next__())
 
+    def append(self, arg):
+        # Create Builder
+        builder = self.io.manager.build(arg)
+
+        # Create ZarrReference
+        ref = self.io._create_ref(builder)
```
rly, line 97: Note: you had `self.io._ZarrIO.__get_ref(builder)` which goes against the convention that private methods are not called outside of its class. This makes refactoring harder. It's totally valid to call this method here, but then I think we should rename it to have a single underscore (and also rename to create instead of get, which implies it is just retrieving an object).

mavaylon1: I kept it double to keep that rigid prevention of others from being able to use it, but given I knew how to get around it then just making the change makes sense.

tests/unit/test_zarrio.py
mavaylon1: Done

rly:
Adding the new check raises this error when calling `read_container.baz_data.data[10]`:
```
if object_path is not None:
            try:
                target_zarr_obj = target_zarr_obj[object_path]
            except Exception:
>               raise ValueError("Found bad link to object %s in file %s" % (object_path, source_file))
E               ValueError: Found bad link to object /new in file /Users/rly/Documents/NWB/hdmf-zarr/test_io.zarr

src/hdmf_zarr/backend.py:741: ValueError
```

@mavaylon1 could you please look into this?

mavaylon1:
@rly This is done.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 0.9.0 (Upcoming)
 ### Enhancements
+* Added support for appending a dataset of references. @mavaylon1 [#203]([https://github.com/hdmf-dev/hdmf-zarr/pull/172](https://github.com/hdmf-dev/hdmf-zarr/pull/203))
```
rly:
```suggestion
* Added support for appending a dataset of references. @mavaylon1 [#203](https://github.com/hdmf-dev/hdmf-zarr/pull/203)
```

src/hdmf_zarr/backend.py
```diff
@@ -440,6 +440,7 @@ def write_builder(self, **kwargs):
         f_builder, link_data, exhaust_dci, export_source, consolidate_metadata = getargs(
             'builder', 'link_data', 'exhaust_dci', 'export_source', 'consolidate_metadata', kwargs
         )
+        # breakpoint()
```
```suggestion
```

src/hdmf_zarr/backend.py
```diff
@@ -519,6 +519,7 @@ def write_group(self, **kwargs):
         parent, builder, link_data, exhaust_dci, export_source = getargs(
             'parent', 'builder', 'link_data', 'exhaust_dci', 'export_source', kwargs
         )
+        # breakpoint()
```
```suggestion
```

src/hdmf_zarr/backend.py
```diff
@@ -629,6 +629,7 @@ def __get_path(self, builder):
         determines the path by constructing it iteratively from the parents of the
         builder.
         """
+        # breakpoint()
```
```suggestion
```

src/hdmf_zarr/backend.py
```diff
@@ -734,6 +734,7 @@ def resolve_ref(self, zarr_ref):
         target_zarr_obj = self.__open_file_consolidated(store=source_file,
                                                         mode='r',
                                                         storage_options=self.__storage_options)
+        # breakpoint()
```
```suggestion
```

src/hdmf_zarr/backend.py
```diff
@@ -742,14 +742,15 @@ def resolve_ref(self, zarr_ref):
...
-    def __get_ref(self, ref_object, export_source=None):
+    def _create_ref(self, ref_object, export_source=None):
         """
         Create a ZarrReference object that points to the given container
 
         :param ref_object: the object to be referenced
         :type ref_object: Builder, Container, ReferenceBuilder
         :returns: ZarrReference object
         """
+        # breakpoint()
```
```suggestion
```

src/hdmf_zarr/backend.py
```diff
@@ -761,7 +761,9 @@ def __get_ref(self, ref_object, export_source=None):
             builder = ref_object.builder
         else:
             builder = self.manager.build(ref_object)
+
         path = self.__get_path(builder)
+        # breakpoint()
```
```suggestion
```

src/hdmf_zarr/zarr_utils.py
```diff
@@ -73,6 +74,16 @@ def __iter__(self):
     def __next__(self):
         return self._get_ref(super().__next__())
 
+    def append(self, arg):
+        # Building the parent first will build the new child.
+        # This correctly sets the parent of the child builder, which is needed to create the reference.
+        self.io.manager.build(arg.parent)
+        builder = self.io.manager.build(arg)
```
rly: On further thought, what if the parent of the new child has not been built? It may be safest to build the whole file then. Can you update the code to do that?

Would this have negative side effects? I can't think of any right now. but maybe if a user adds a container to the file, appends a reference elsewhere in the file, which would trigger a full build, then tries to remove the container or its parent from the file - the builder for the removed container will already have been built and attached. This is why usually all the building and reference-making happen during write() after which the user is not supposed to modify the in-memory objects anymore. 

I wonder if it would be better to defer the full file building to write() if that is possible, like we could store a proxy object here that gets processed during the io.write() call. Can you add a comment about this case in the code or create an issue ticket? We could try to address this now but for the sake of getting this out the door, I say we move forward and just do a full-file build here.

mavaylon1: "On further thought, what if the parent of the new child has not been built? It may be safest to build the whole file then. Can you update the code to do that?"

I'm not sure what you mean. Why wouldn't the parent of the new child not be built? Also building the entire file for a single append doesn't seem efficient. 

I agree with the idea of making an issue ticket for the proxy idea.

mavaylon1:
@rly I made the changes. The main thing is to take a look at my notes. Your comment "the parent of the new child has not been built" is weird since the prior code said to build the parent. I assumed you meant the parent of the child could not be the root, leaving the path to the child being possibly incorrect if nested. To ensure a proper path to the child for the reference, we need to build the root. Let me know if that is what you meant. 

I added a loop to do so. I did not add a nested test. If we really want one we can do so in a followup PR, but I think I want to explore the proxy idea in case that replaces this.

rly (approved):

PR #206, opened 2024-07-18
S3 test

### LINKED ISSUE #185: [Bug]: Remote read with/without consolidated metadata is not being tested
### What happened?

https://github.com/hdmf-dev/hdmf-zarr/pull/184 adds a bug fix for reading files from S3 with consolidated metadata. However, we currently are not testing reading with consolidated metadata only for local files. Since consolidated metadata is particularly relevant for reading from S3, it would be useful to add tests to ensure this is working correctly with files on S3 as well. 
 

### Steps to Reproduce

I believe we currently do no have a test file with consolidated metadata on DANDI. It would be useful to add some small Zarr files with and without consolidated metadata to our test DANDI dataset so that we can test that the remote files are being opened correctly in the streaming tests. Currently we are using a dataset from the Allen Institute for the streaming tests but it would be useful to test the behavior for streaming access more thoroughly using our own files. 

To test whether a file is indeed opened correctly with consolidated metadata we can use the following check in the unit tests :
https://github.com/hdmf-dev/hdmf-zarr/blob/afaab5fbca03a0979d9f234e13e43027f545c366/tests/unit/test_zarrio.py#L160-L166

### Traceback

```shell
N/A
```

### Operating System

macOS

### Python Executable

Conda

### Python Version

3.9

### Package Versions

_No response_

### Code of Conduct

mavaylon1:
## Motivation
Fix #185 

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/tests/unit/test_fsspec_streaming.py
+++ b/tests/unit/test_fsspec_streaming.py
@@ -2,6 +2,8 @@ import unittest
 from hdmf_zarr import NWBZarrIO
 from .utils import check_s3fs_ffspec_installed
 
+import zarr
+
 HAVE_FSSPEC = check_s3fs_ffspec_installed()
 
 
@@ -24,3 +26,10 @@ class TestFSSpecStreaming(unittest.TestCase):
         self.assertEqual(len(nwbfile.electrode_groups), 2)
         self.assertEqual(len(nwbfile.electrodes), 1152)
         self.assertEqual(nwbfile.institution, "AIND")
+
+    @unittest.skipIf(not HAVE_FSSPEC, "fsspec not installed")
+    def test_s3_open_with_consolidated_(self):
+        s3_path = "https://dandiarchive.s3.amazonaws.com/zarr/ccefbc9f-30e7-4a4c-b044-5b59d300040b/"
+        with NWBZarrIO(s3_path, mode='r') as read_io:
+            read_io.open()
+            self.assertIsInstance(read_io.file.store, zarr.storage.ConsolidatedMetadataStore)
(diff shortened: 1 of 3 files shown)


tests/unit/test_fsspec_streaming.py
```diff
@@ -24,3 +26,10 @@ def test_fsspec_streaming(self):
         self.assertEqual(len(nwbfile.electrode_groups), 2)
         self.assertEqual(len(nwbfile.electrodes), 1152)
         self.assertEqual(nwbfile.institution, "AIND")
+
+    @unittest.skipIf(not HAVE_FSSPEC, "fsspec not installed")
+    def test_s3_open_with_consolidated_(self):
+        s3_path = "https://dandiarchive.s3.amazonaws.com/zarr/ccefbc9f-30e7-4a4c-b044-5b59d300040b/"
```
stephprince, line 35: Can you add a short comment explaining what this links to / how it is different from the path provided in the `test_fsspec_streaming` test?

tests/unit/test_fsspec_streaming.py
```diff
@@ -24,3 +26,10 @@ def test_fsspec_streaming(self):
...
+        with NWBZarrIO(s3_path, mode='r') as read_io:
+            read_io.open()
+            self.assertIsInstance(read_io.file.store, zarr.storage.ConsolidatedMetadataStore)
```
stephprince, line 38: The original issue mentions also adding a test to force a regular open without consolidated metadata - is there a reason that's not included for the fsspec streaming?

```python
# Confirm that opening the file IN 'r-' mode indeed forces a regular open without consolidated metadata
with NWBZarrIO(s3_path, mode='r-') as read_io:
    read_io.open()
    self.assertIsInstance(read_io.file.store, zarr.storage.DirectoryStore)
```

mavaylon1: I don't think the original PR was asking for that. I understood it as "like this". The ticket does explicitly say open a file from dandi, but I don't see a reason why not to add your suggestion. Can do.

tests/unit/test_fsspec_streaming.py
mavaylon1: I think the name of the test is sufficient, but sure.

stephprince (approved):

PR #240, opened 2024-11-22
Remove Python 3.8 and add Python 3.13

### LINKED ISSUE #234: [Feature]: Remove support for python 3.8

Python 3.8 is deprecated. Update the pyproject.toml

_No response_

Above

Yes.

### Code of Conduct

mavaylon1:
Remove Python 3.8 and add Python 3.13 
Fix #234 

## Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -62,11 +62,11 @@ body:
     attributes:
       label: Python Version
       options:
-        - "3.8"
         - "3.9"
         - "3.10"
         - "3.11"
         - "3.12"
+        - "3.13"
     validations:
       required: true
   - type: textarea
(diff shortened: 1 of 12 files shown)


rly:
Parallel write tests are failing on Windows with Python 3.13. Not sure why. I also ran all tests on this branch to see if it is only Python 3.13: https://github.com/hdmf-dev/hdmf-zarr/actions/runs/11968032434

If it is only Python 3.13 and there is no easy fix, I propose, in the interest of our schedules, that we remove python 3.13 support (just remove the line from `pyproject.toml`) for now, and we can add it in a later version of hdmf-zarr. We can keep the tests and CI as they are while we work through the issue.

mavaylon1:
@rly If I understand what you are saying, keep the testing with 3.13 but remove support for 3.13 for users via pyproject.toml. To do that I think we have add a <3.13, not just comment out the classifier part. I believe that is just metadata.

rly:
The Windows Python 3.13 error was fixed in pytest 8.1.0: https://github.com/pytest-dev/pytest/pull/11875

I think updating pytest and related dependencies should resolve the error (I just pushed changes for that to this branch). We should update all the fixed testing dependencies before the 1.0 release anyway.

rly (approved):

PR #256, opened 2025-01-08
Fix relative references when store not in cwd

### LINKED ISSUE #255: [Bug]:
### What happened?

I am trying to open a zarr nwb in order to append CCF information to the electrodes table, but it fails to open.

Note that .zmetadata is present in the Zarr NWB file. It's worth mentioning that I generated this hdmf-zarr==0.8.0 but and now trying to open it with latest (0.10.0), perhaps that is causing this. 

### Steps to Reproduce

```shell
Run the following code snippet, where there is a Zarr NWB present at `scratch_nwb_path`


    print('Reading NWB in append mode:', scratch_nwb_path)
    print(os.listdir(scratch_folder))
    if scratch_nwb_path.is_dir():
        print(os.listdir(scratch_nwb_path))
    with NWBZarrIO(str(scratch_nwb_path), mode='a') as read_io:
        nwb = read_io.read()
```

### Traceback

```shell
Reading NWB in append mode: ../scratch/ecephys_747816_2024-08-14_13-57-37_experiment1_recording1.nwb
['ecephys_747816_2024-08-14_13-57-37_experiment1_recording1.nwb']
['file_create_date', 'intervals', 'analysis', 'specifications', 'stimulus', '.zattrs', 'general', 'acquisition', '.zmetadata', '.zgroup', 'processing', 'timestamps_reference_time', 'identifier', 'units', 'session_description', 'session_start_time']
  File "/opt/conda/lib/python3.9/site-packages/hdmf_zarr/backend.py", line 566, in __open_file_consolidated
    return zarr.open_consolidated(
  File "/opt/conda/lib/python3.9/site-packages/zarr/convenience.py", line 1360, in open_consolidated
    meta_store = ConsolidatedStoreClass(store, metadata_key=metadata_key)
  File "/opt/conda/lib/python3.9/site-packages/zarr/storage.py", line 3046, in __init__
    meta = json_loads(self.store[metadata_key])
  File "/opt/conda/lib/python3.9/site-packages/zarr/storage.py", line 1120, in __getitem__
    raise KeyError(key)
KeyError: '.zmetadata'

During handling of the above exception, another exception occurred:

  File "/root/capsule/code/run_capsule.py", line 140, in <module>
    if __name__ == "__main__": run()
  File "/root/capsule/code/run_capsule.py", line 124, in run
    nwb = read_io.read()
  File "/opt/conda/lib/python3.9/site-packages/hdmf/utils.py", line 668, in func_call
    return func(args[0], **pargs)
  File "/opt/conda/lib/python3.9/site-packages/hdmf/backends/io.py", line 56, in read
    f_builder = self.read_builder()
  File "/opt/conda/lib/python3.9/site-packages/hdmf/utils.py", line 668, in func_call
    return func(args[0], **pargs)
  File "/opt/conda/lib/python3.9/site-packages/hdmf_zarr/backend.py", line 1470, in read_builder
    f_builder = self.__read_group(self.__file, ROOT_NAME)
  File "/opt/conda/lib/python3.9/site-packages/hdmf_zarr/backend.py", line 1548, in __read_group
    sub_builder = self.__read_group(sub_group, sub_name)
  File "/opt/conda/lib/python3.9/site-packages/hdmf_zarr/backend.py", line 1548, in __read_group
    sub_builder = self.__read_group(sub_group, sub_name)
  File "/opt/conda/lib/python3.9/site-packages/hdmf_zarr/backend.py", line 1557, in __read_group
    self.__read_links(zarr_obj=zarr_obj, parent=ret)
  File "/opt/conda/lib/python3.9/site-packages/hdmf_zarr/backend.py", line 1576, in __read_links
    target_name, target_zarr_obj = self.resolve_ref(link)
  File "/opt/conda/lib/python3.9/site-packages/hdmf_zarr/backend.py", line 822, in resolve_ref
    target_zarr_obj = self.__open_file_consolidated(
  File "/opt/conda/lib/python3.9/site-packages/hdmf_zarr/backend.py", line 573, in __open_file_consolidated
    return zarr.open(
  File "/opt/conda/lib/python3.9/site-packages/zarr/convenience.py", line 137, in open
    raise PathNotFoundError(path)
zarr.errors.PathNotFoundError: nothing found at path ''
(base) root@a09d29db9803:~/capsule/code# 
```
```


### Operating System

Windows

### Python Version

3.9

### Package Versions

_No response_

### PR DESCRIPTION
**Pull request by rly**
## Motivation

Fix #255. Slight rework of how references are handled. 

Currently, the "source" of links and references are relative to the directory that **contains** the currently open Zarr store.

Internal Zarr links and references use the name of the Zarr store as the "source". For example, if `file1.zarr` contains an internal link, that link looks like:
```json
{
    "name": "foo_link",
    "path": "/buckets/bucket1/foo_holder/foo1",
    "source": "file1.zarr"
}
```

External links: If `file0.zarr` and `file1.zarr` are in the same directory, and `file1.zarr` has an external link to a group in `file0.zarr`, that link looks like:
```json
{
    "name": "foo_link",
    "path": "/buckets/bucket1/foo_holder/foo1",
    "source": "file0.zarr"
}
```

When reading `file1.zarr` from elsewhere, e.g., `ZarrIO("a/b/c/file1.zarr", mode="r")`, we need to make sure that the relative paths are resolved correctly from `a/b/c`. Approach 1: Fix this.

An alternative approach, which I believe was the original implementation, is that instead of writing paths relative to the directory that contain the currently open Zarr store, write paths relative to the currently open Zarr store. So internal links and references would have the `source: "."` and the above external link example would have the `source: "../file0.zarr"`. Approach 2: Return to this pre-0.10.0 way of storing "source" in links and references.

I propose we take approach 2, because otherwise links in files written under the original implementation would not be readable - I tested this and indeed files with links written in 0.8.0 cannot be read in 0.10.0 - and approach 2 does not suffer any disadvantages in my eyes.

@mavaylon1 Let me know what you think.

## TODO

- [ ] Update tests once we settle on an approach

## Checklist


mavaylon1:
We got rid of "." notation because there was no logic to differentiate internal and external links. I don't see the issue within your example. Are you saying that if we moved file1 the path to file0 will be broken? If so then yes it would. We wouldn't want people moving things around relative to each other. 

I don't agree with going back to the idea pre-0.10.0, the changes were made to address issue tickets for external links and resolving references properly. I'd prefer amending the system if there are holes, but reverting back will just bring back old problems. 

As for "." notation, I am not against changing the system to support that in some way, though I do not see an issue with the explicit naming in your example.

mavaylon1:
> We got rid of "." notation because there was no logic to differentiate internal and external links. I don't see the issue within your example. Are you saying that if we moved file1 the path to file0 will be broken? If so then yes it would. We wouldn't want people moving things around relative to each other.
> 
> I don't agree with going back to the idea pre-0.10.0, the changes were made to address issue tickets for external links and resolving references properly. I'd prefer amending the system if there are holes, but reverting back will just bring back old problems.
> 
> As for "." notation, I am not against changing the system to support that in some way, though I do not see an issue with the explicit naming in your example.

I see the issue #255 let me think on this.

rly:
The team @mavaylon1 @oruebel and I discussed this PR today.

We had changed how links and references are stored and read because of issues with how external links are stored/read pre-0.10.0. @mavaylon1 will provide more details separately.

One strong reason for writing "." instead of the explicit filename of the Zarr store for internal links and references is that if the user renames the Zarr store, then they would have to update all internal links and references with the new name; otherwise, those would all break. 

We agreed that because of this reason, we will revert to writing "." for internal links and references and use the currently open Zarr store as the base directory from which external links are relative.

I will update the tests to reflect this.

.pre-commit-config.yaml
```diff
@@ -8,9 +8,6 @@ repos:
     -   id: check-added-large-files
     -   id: check-json
     -   id: check-toml
-    -   id: name-tests-test
```
rly, line 11: These tests were failing and not necessary for this repo

src/hdmf_zarr/backend.py
```diff
@@ -726,55 +726,16 @@ def __get_path(self, builder):
             path = "%s%s" % (delim, delim.join(reversed(names)))
         return path
 
-    @staticmethod
```
rly, line 729: This method was convoluted and not really used internally. I have simplified the code and its usage. This introduces a breaking change, but since we have not released 1.0 and I doubt users are using this method, I think it is OK to remove.

src/hdmf_zarr/backend.py
```diff
@@ -726,55 +726,16 @@ def __get_path(self, builder):
...
         :param zarr_object: Object for which we are looking up the path
         :type zarr_object: Zarr Group or Array
         :return: String with the path
         """
-        filepath, objectpath = ZarrIO.get_zarr_paths(zarr_object)
-        parentpath = os.path.dirname(objectpath)
-        return parentpath
-
-    @staticmethod
-    def is_zarr_file(path):
```
rly, line 767: Same as above, except this method isn't used internally anywhere.

tests/unit/base_tests_zarrio.py
```diff
@@ -1029,11 +1029,11 @@ def test_basic(self):
                 export_io.export(src_io=read_io)
 
         self.assertTrue(os.path.exists(self.store_path[1]))
-        self.assertEqual(os.path.relpath(foofile.container_source), self.store_path[0])
```
rly, line 1032: Calls to `os.path.relpath` rely on the current working directory. We should avoid using that so that we know the code works regardless of where the code is run. Also in general I think we should avoid modifying the value being tested and instead modify what it is tested against because the expected value is clearer.

mavaylon1 (approved):

Issue #285, opened 2025-07-30
Discussion: Change default codec for hdmf_zarr_utils.ZarrReference


ehennestad:
From documentation:

> By default, [ZarrIO](https://hdmf-zarr.readthedocs.io/en/stable/hdmf_zarr.backend.html#hdmf_zarr.backend.ZarrIO) uses the numcodecs.pickles.Pickle codec to encode object references defined as py:class:~hdmf_zarr.utils.ZarrReference dicts in datasets. Users may set the codec used to encode objects in Zarr datasets via the object_codec_class parameter of the __init__() constructor of [ZarrIO](https://hdmf-zarr.readthedocs.io/en/stable/hdmf_zarr.backend.html#hdmf_zarr.backend.ZarrIO). E.g., we could use ZarrIO( ... , object_codec_class=numcodecs.JSON) to serialize objects using JSON.

I am working on adding read support for Zarr files in MatNWB. It seems to me that using the numcodecs.pickles.Pickle codec for storing `hdmf_zarr.utils.ZarrReference` objects adds extra hurdles for reading zarr files with other languages than Python. 

I wonder if it would be better to use numcodecs.JSON by default, and maybe even remove the option to set a codec via parameter?

bendichter:
@rly would be interested in your thoughts on this. Is it a priority to ensure these files are readable by other languages?

oruebel:
> I wonder if it would be better to use numcodecs.JSON by default, and maybe even remove the option to set a codec via parameter?

At least for NWB, I think choosing one codec would be useful. It's been a while since I've hacked on the codec part of hdmf-zarr but as far as I recall we had used the `pickle` codecmainly because it is the most broad and reliable in terms of its ability to serialize Python objects. The `JSON` codec had issues with serializing certain data types and using pickle was the simplest solution at the time. I don't recall which specific data types where causing issues, but I  think structured arrays was one of them. However, a lot of things have changed since then so its worth re-evaluating that choice. 

As a first step, I think it would be useful to understand whether the `JSON` codec supports serializing all the types we need. Changing the codec is a fairly minimal code change so I created a draft PR to test this here https://github.com/hdmf-dev/hdmf-zarr/pull/286 so we can see which tests will fail when using `JSON` as the codec.

It looks like there are a few errors that happen in the `JSON` codec:

* `E           ValueError: setting an array element with a sequence. The requested array would exceed the maximum number of dimension of 1.`
[...]

PR #286, opened 2025-07-31
Test if changing default codec to JSON would work

oruebel:
## Motivation

See #285 The goal is to see if `JSON` codec works with the types that we need to support. 

- [X] Change default codec in `ZarrIO` to JSON
- [ ] Remove the `object_codec_class` option from `NWBZarrIO` 

## Checklist

- [ ] Does the PR clearly describe the problem and the solution?
- [ ] Does the PR use "Fix #XXX" notation to tell GitHub to close the relevant issue numbered XXX when the PR is merged?


oruebel:
Superseded by the Zarr v3 updates in #325

PR #294, opened 2025-08-11
Add ZarrIO._file and fix load_namespaces

### LINKED ISSUE #293: [Bug]: PyNWB validation fails on Zarr data: 'NWBZarrIO' object has no attribute '_file'
### What happened?

Validating a Zarr NWB file with pynwb fails. Related to https://github.com/NeurodataWithoutBorders/pynwb/issues/2119

As a quick fix, I suggest we add attribute `_file` while the other changes in https://github.com/hdmf-dev/hdmf/pull/1299 and https://github.com/hdmf-dev/hdmf-zarr/pull/292 are gradually merged (the HDMF change is a major breaking change).

### Steps to Reproduce

```shell
with hdmf_zarr.NWBZarrIO(path=nwbfile_path, mode="r") as io:
    invalidations = pynwb.validate(io=io)
```

### Traceback

```shell
  File "<stdin>", line 2, in <module>
  File "/Users/rly/Documents/NWB/hdmf/src/hdmf/utils.py", line 582, in func_call
    return func(**pargs)
           ^^^^^^^^^^^^^
  File "/Users/rly/Documents/NWB/pynwb/src/pynwb/validation.py", line 165, in validate
    validation_errors = _validate_single_file(path=path, **kwargs)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rly/Documents/NWB/pynwb/src/pynwb/validation.py", line 183, in _validate_single_file
    cached_namespaces, manager, namespace_dependencies = get_cached_namespaces_to_validate(path=path, 
                                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rly/Documents/NWB/pynwb/src/pynwb/validation.py", line 66, in get_cached_namespaces_to_validate
    file=io._file)
         ^^^^^^^^
AttributeError: 'NWBZarrIO' object has no attribute '_file'. Did you mean: 'file'?
```

### Operating System

macOS

### Python Version

3.13

### Package Versions

_No response_

rly:
## Motivation

Fix #293.

~Need to determine how to validate string/isodatetime arrays in the HDMF validator. Will likely require changes in HDMF.~ This will be handled in a separate PR.

## How to test the behavior?
```
pytest
```

## Checklist


src/hdmf_zarr/backend.py
```diff
@@ -167,6 +167,15 @@ def file(self):
         """
         return self.__file
 
+    @property
+    def _file(self):
```
oruebel, line 162: Instead of adding this property, should `def file` be renamed instead? At the very least, I would recommend that this property should `return self.file` instead to make sure that any logic that may be added later doesn't get duplicated or that the two functions don't get out-of-sync.

rly: Ideally `io._file` is not called by pynwb. But I agree. Here, let's rename `def file` to `def _file` for now. `HDF5IO` does not publicly expose the `h5py.File` object -- all operations should go through `HDF5IO` to keep the two in sync. For the same reason, I would say `ZarrIO` should not expose the private open group. What do you think?

oruebel: > all operations should go through `HDF5IO` to keep the two in sync.

Doe you mean `HDMFIO` here?

> `HDF5IO` does not publicly expose the `h5py.File` object

I think `HDF5IO` has the same `_file` property 

> For the same reason, I would say `ZarrIO` should not expose the private open group.

I think exposing them as a "private" `_file` property that is intended for internal use is fine. At the very least this is useful for testing. However, I agree that it should not be a public property.

stephprince (approved):

rly:
macos-python3.9-minimum tests are failing due to a conflict with numcodecs being tracked in #297

PR #322, opened 2025-12-24
Ignore specifications group when reading root builder

### LINKED ISSUE #321: [Bug]: Read builder contains specifications group
### What happened?

In HDF5IO, `read_builder` does not read the `specifications` group. This change was made in May 2018: https://github.com/hdmf-dev/hdmf/commit/8e552932d7f79827d3ac5598e7f0de5d1c0277ae

In ZarrIO, `read_builder` reads the `specifications` group. @oruebel do you happen to recall why this was omitted when creating `ZarrIO`? Was the initial implementation before May 2018?

For a test file, listing the root builder's keys shows:
```
['acquisition', 'analysis', 'general', 'processing', 'specifications', 'stimulus', 'units', 
'file_create_date', 'identifier', 'session_description', 'session_start_time', 
'timestamps_reference_time', '.specloc', 'namespace', 'neurodata_type', 'nwb_version', 'object_id']
```

This inconsistency would normally not be a problem because I think extra objects in the builder that don't belong to a container are ignored, but the specifications JSON string datasets are written with parameters `dtype=object, object_codec=numcodecs.JSON()` and `zarr_dtype` is set to "scalar". In https://github.com/hdmf-dev/hdmf-zarr/pull/310, we are getting rid of `zarr_dtype="scalar"`, and #310 proposes to set the `zarr_dtype` to "object". But `zarr_dtype="object"` is reserved for object references. On read, because the scalar JSON string datasets are read during `read_builder`, this is all messed up in `read_dataset` unless you add special code in #310 which copilot added.

I think we should resolve this issue by not reading the specifications in `ZarrIO.read_builder` to begin with and not writing a `zarr_dtype`. I would also like to make `ZarrIO` and `HDF5IO` as consistent as possible to ease maintenance.

### Steps to Reproduce

```shell
Load #310 and remove the `dtype != "object"` case in the new conditional:

# Read scalar dataset immediately for simple types (not object/reference types).
# Old style scalars have dtype == "scalar" with shape (1,), while new style scalars  
# have shape () with actual dtype stored in zarr_dtype. Object-dtype scalars (e.g., 
# spec JSON data) and reference datasets should remain as lazy arrays for proper handling.
if not is_ref_dataset and (dtype == "scalar" or (zarr_obj.shape == () and dtype != "object")):
    data = zarr_obj[()]
```

### Traceback

```shell
FAILED tests/unit/test_zarrio.py::TestExportZarrToZarrDirectoryStore::test_basic - UnboundLocalError: cannot access local variable 'builder' where it is not associated with a value

in `ZarrIO._create_ref` because '{' is passed as the ref_object.
```

### Operating System

macOS

### Python Version

3.14

### Package Versions

_No response_

rly:
## Motivation

Fix #321 

## How to test the behavior?
```
Run tests
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Fixed zarr array `.info` property to display compression data ("No. bytes stored" and "Storage ratio") when using consolidated metadata stores by patching `ConsolidatedMetadataStore.getsize()` to query the underlying chunk store. @rly [#305](https://github.com/hdmf-dev/hdmf-zarr/pull/305)
 - Fixed bug where the cached spec was not included in consolidated metadata when exporting using `ZarrIO`. @rly [#315](https://github.com/hdmf-dev/hdmf-zarr/pull/315)
 - Fixed bug where consolidated metadata was always written by `ZarrIO`. @rly [#315](https://github.com/hdmf-dev/hdmf-zarr/pull/315)
+- Fixed bug where the `specifications` group and its contents were read during `ZarrIO.read_builder`. @rly [#322](https://github.com/hdmf-dev/hdmf-zarr/pull/322)
 
 ### Changed
 - Updated documentation about how the `zarr_dtype` attribute is used to specify a compound dtype (structured array) for a dataset. @rly [#313](https://github.com/hdmf-dev/hdmf-zarr/pull/313)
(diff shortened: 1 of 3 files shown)


rly:
In the interest of keeping the code simple and maintainable - instead of the versatile approach of adding an `ignore` argument to `read_group` at every level of recursion when iterating through the file, since we currently only want to ignore just the `specifications` group, `ignore` is passed to `read_group` only for the root group.

oruebel (approved):

Issue #309, opened 2025-12-13
[Feature]: Write scalar datasets with shape ()


rly:
### What would you like to see added to HDMF-ZARR?

See https://github.com/hdmf-dev/hdmf-zarr/issues/284 and 
https://github.com/hdmf-dev/hdmf-zarr/blob/8ee50283b192931000d7946e170747882f6c5c16/src/hdmf_zarr/backend.py#L1504

Zarr v2 supports writing arrays with shape None (or `()`). This is consistent with `HDF5IO`: 
https://github.com/hdmf-dev/hdmf/blob/a90f8ed0588fcdd84d4af1d693e851099ced6066/src/hdmf/backends/hdf5/h5tools.py#L1278

We should do the same with `ZarrIO`. Doing this will remove a special handling case, and remove a deviation between `HDF5IO` and `ZarrIO` as we look to merge it into core HDMF.

### Is your feature request related to a problem?

This would simplify the resolution of #277.

### What solution would you like?

Use shape () and stop writing the attribute `zarr_dtype=scalar` but keep processing it on read in case it is present. 

### Do you have any interest in helping implement the feature?

Yes.

PR #325, opened 2026-02-24
Migrate hdmf-zarr from zarr-python v2 to v3

### LINKED ISSUE #202: [Feature]: Support zarr-python v3

Zarr-Python is moving closer to a v3 release which has significant changes in the API that result in errors in hdmf-zarr. It is not clear how stable the current alpha release is right now. But we should add support for Zarr Python v3 soon. See #200 for more details. 

_No response_

^

Yes, but I would need guidance.

### Code of Conduct

bendichter:
## Summary

Migrate hdmf-zarr to zarr-python v3. `ZarrIO` and `NWBZarrIO` write and read Zarr v3 format only. Legacy Zarr v2 files are read with the new read-only `ZarrV2IO` and `NWBZarrV2IO` backends, which can also convert them to v3 (`NWBZarrV2IO.export_to_v3` / `convert_to_v3`).

Closes #202

### Highlights

- Adopts the unified Zarr v3 storage convention shared with Zindi/LINDI (`_LINKS`, `_DTYPE`, `_REFERENCE_FIELDS`, zero-dimensional scalar datasets, JSON references). Zindi-generated stores can be read with `NWBZarrIO`.
- `ZarrDataIO` follows the zarr v3 codec API (`compressors`, `serializer`, `filters`) and adds a `shards` argument, including automatic sharding.
- Supports `LocalStore`, `FsspecStore`, and any zarr v3 `Store` subclass. Remote writes are not supported.
- Removes the `synchronizer` and `object_codec_class` arguments.
- Raises minimums: Python >=3.12, `zarr>=3.3.0`, `hdmf>=6.2.0`, `numpy>=2.0.0`, `pynwb>=4.2.0`.

See the CHANGELOG for the full list of changes and the related PRs merged into this branch.


oruebel:
@bendichter thanks for taking a crack at this. I'll take a look. 

Do you know if this is compatible with reading existing Zarr v2 files? I think it is fine to drop Zarr v2 support for writing, but it would be great to have a pathway to maintain read for existing v2 data.

src/hdmf_zarr/nwb.py
```diff
@@ -38,6 +39,15 @@ def __init__(self, **kwargs):
...
         )
 
+        # Emit deprecation warning for synchronizer if provided
+        if synchronizer is not None and synchronizer is not False:
+            warnings.warn(
+                "The 'synchronizer' parameter is deprecated and ignored in zarr v3. "
+                "Zarr v3 does not support synchronizers.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
oruebel: Since Zarr v3 would be a major release, I think it should be fine to just remove the `synchronizer` argument if Zarr v3 doesn't support it.

src/hdmf_zarr/utils.py
```diff
@@ -406,6 +406,36 @@ def read_namespace(self, ns_path):
...
+            if not name.startswith("_")
+        ] if isinstance(cls, type)}
+        # Map from numcodecs codec_id to wrapper class name
+        codec_class_name = type(codec).__name__
+        if codec_class_name in wrapper_map:
+            import warnings
+            with warnings.catch_warnings():
+                warnings.filterwarnings("ignore", message="Numcodecs codecs are not in the Zarr")
+                return wrapper_map[codec_class_name](**config)
+    raise TypeError(f"Cannot convert {type(codec)} to a zarr v3 codec")
```
oruebel: @bendichter at first glance, it seems that this used in ZarrDataIO to ensure compatibility with how Zarr v2 was configuring codes. Since this is only for write and since this would be a major release, I think it would be fine to require `ZarrDataIO` to accept only configurations that are allowed with Zarr v3.

src/hdmf_zarr/zarr_utils.py
```diff
@@ -106,12 +89,6 @@ def __next__(self):
 
     def append(self, arg):
         # Building the root parent first.
-        # (Doing so will correctly set the parent of the child builder, which is needed to create the reference)
-        # Note: If the arg is a nested child such that objB is the parent of arg and objA is the parent of objB
-        # (and objA is not the root), then we need to have objA already added to the root as a child. Otherwise,
-        # the loop will use objA as the root. This might not raise an error (meaning the path could be correct),
-        # but it could lead to having an incorrect path for the reference.
-        # Having objA NOT be an orphaned container ensures correct functionality.
```
oruebel, line 114: I think this removed comment still applies

src/hdmf_zarr/zarr_utils.py
```diff
@@ -173,17 +152,18 @@ def __init__(self, **kwargs):
...
         self.__types = types
         tmp = list()
         for i in range(len(self.dataset.dtype)):
             sub = self.dataset.dtype[i]
             if np.issubdtype(sub, np.dtype("O")):
                 tmp.append("object")
-            if sub.metadata:
+            elif np.issubdtype(sub, np.str_):
+                # In zarr v3, string fields in compound dtypes use fixed-length Unicode
+                # Check if this field holds JSON-serialized references
```
oruebel, line 170: @bendichter I haven't looked at `backend.py` yet, but how is this the fixed-length string handled to avoid potentially cutting of a JSON reference. I think this is mainly relevant for `TimeSeriesRefenceVectorData` which is a compound type of `(int, int, reference)`

src/hdmf_zarr/nwb.py
bendichter: Done - removed the `synchronizer` parameter entirely from both `ZarrIO` and `NWBZarrIO` instead of deprecating it.

src/hdmf_zarr/utils.py
bendichter: Done - removed `_numcodec_to_zarr_v3` and updated `ZarrDataIO` to accept only zarr v3 codecs (`BytesBytesCodec` for compressors, `ArrayArrayCodec` for filters). Also removed the `object_codec_class` parameter from `ZarrIO` entirely (same logic). Updated `hdf5_to_zarr_filters` to produce zarr v3 wrapped codecs directly via `zarr.codecs.numcodecs`.

src/hdmf_zarr/zarr_utils.py
bendichter: Good catch - restored the comment.

bendichter:
@oruebel Thanks for the review! Here's a summary of the changes in the latest commit:

**Re: zarr v2 read compatibility** - zarr-python v3 can transparently read zarr v2 format files as long as they don't use `object_codec` (pickle/JSON codecs for object dtype arrays). Files with `object_codec` (which hdmf-zarr v2 used for references and specs) cannot be read by zarr v3 directly - those would need a migration tool. The vast majority of simple NWB-Zarr files should read fine. I've skipped the fsspec streaming tests since the S3 test files use v2 object_codec format.

**Changes in this commit addressing your review:**
1. **Removed `synchronizer` parameter** entirely from `ZarrIO` and `NWBZarrIO` (not just deprecated)
2. **Removed `object_codec_class` parameter** entirely from `ZarrIO` (same logic)
3. **Removed `_numcodec_to_zarr_v3` compat wrapper** - `ZarrDataIO` now accepts only zarr v3 codecs (`BytesBytesCodec` for compressors, `ArrayArrayCodec` for filters). Users should use `zarr.codecs` (e.g., `zarr.codecs.BloscCodec()`) or `zarr.codecs.numcodecs` wrappers instead of raw numcodecs objects.
4. **Updated `hdf5_to_zarr_filters`** to produce zarr v3 wrapped codecs directly
5. **Restored removed comment** in `zarr_utils.py` about orphaned containers

src/hdmf_zarr/zarr_utils.py
bendichter: Good catch! The hardcoded `U256` was indeed a truncation risk. A typical JSON reference is ~209 chars, but deeply nested paths could easily exceed 256.

Fixed in b3ccee0 — both compound dtype code paths (`__write_dataset__` for refs and `__list_fill__` for regular compound types) now dynamically compute the max string length from the actual data before building the numpy dtype. This ensures the fixed-length Unicode field is always sized to fit.

oruebel:
> Files with `object_codec` (which hdmf-zarr v2 used for references and specs) cannot be read by zarr v3 directly - those would need a migration tool. The vast majority of simple NWB-Zarr files should read fine. I

Just to clarify, does Zarr v3 fail to read 1) any v2 file that contains `object_codec` datasets or  2) "just" not support reading the respective datasets within the v2 file? 

If it is 1) then I think that would mean that no NWB Zarr file would be readable, because of the cached spec. I think in that case we may need some form of a migration tool.  

If it is 2) then I think the main problem cases are:
* spec JSON dataset
* compound dataset with variable-length strings
* compound dataset with references (which are encoded a variable-length JSONs)
I am wondering if those cases could be handled here?

bendichter:
what NWB datasets currently use Zarr v2?

bendichter:
I tested this systematically. It's scenario 2 — zarr v3 fails on individual `object_codec` datasets, not the entire file. However, the practical impact is closer to 1, because `.groups()` and `.arrays()` iterate all children and fail if *any* child has `object_codec`:

```python
# These work fine:
root = zarr.open_group(store, mode='r')  # opening the group succeeds
root['normal_data'][:]                    # accessing non-object arrays by name works
root['subgroup']                          # accessing subgroups by name works

# These fail:
root['spec_dataset']    # accessing an object_codec array by name fails
root.arrays()           # iterating arrays fails if ANY array has object_codec
root.groups()           # iterating groups also fails (same underlying iteration)
root.members()          # same
```

Since hdmf-zarr's `read_builder` uses `.groups()` and `.arrays()` to walk the hierarchy, and every NWB-Zarr v2 file has `object_codec` datasets in `/specifications`, no existing NWB-Zarr file can be read with this branch as-is.

We have a few options:
1. **Skip for now** — v2 read support is a separate concern from the v3 write migration. A migration tool or a separate `read_v2` pathway could be added later.
2. **Replace `.arrays()`/`.groups()` with filesystem-level listing** — manually scan for `.zarray`/`.zgroup` files, catch errors on individual arrays, and handle object_codec datasets specially (read the raw JSON chunks and decode them ourselves).
3. **Try-catch during iteration** — wrap individual array access in try/except and log warnings for unreadable datasets.

I think option 1 is reasonable for this PR since it's already a major release. We could file a follow-up issue for v2 read compatibility. What do you think?

oruebel:
> what NWB datasets currently use Zarr v2?

@bendichter Any of the Allen Institute for Neural Dynamics (AIND) datasets should be Zarr v2. However, I am not sure if they have published those on DANDI. If there are no NWB Zarr files on DANDI, then I think the migration problem may be mainly an AIND-specific issue.

> 1. **Skip for now** — v2 read support is a separate concern

I agree that this doesn't necessarily need to be part of this PR. However, we should have some answer for the migration question for the major release.

bendichter:
@oruebel I know AIND is one of the main groups pushing for Zarr v3, so I think from their perspective they would rather us push forward with this

test_gallery.py
```diff
@@ -81,6 +81,11 @@ def _import_from_file(script):
     "__array__ must implement 'dtype' and 'copy' keyword arguments."
 )
 
+_zarr_numcodecs_warning_re = r"Numcodecs codecs are not in the Zarr version 3 specification.*"
+_zarr_consolidated_warning_re = r"Consolidated metadata is currently not part in the Zarr format 3 specification.*"
+_zarr_unstable_dtype_warning_re = r"The data type .* does not have a Zarr V3 specification.*"
+_hdmf_dtype_conversion_warning_re = r"Spec '.*': Value with data type .* is being converted to data type .*"
```
oruebel, line 87: Some of these look like they would be important warnings. Do you know which cases cause these warnings and if they can indeed be safely ignored. Also, how often do these warnings occur? We can ignore them in the galleries to ensure clean rendering of tutorials, but if these warnings occur frequently then it would be useful to handle them to avoid that warnings are issues for every file.

bendichter: Good question. Here's a breakdown:

**1. "Numcodecs codecs are not in the Zarr version 3 specification"** (UserWarning) — Fires every time a numcodecs codec is instantiated, so very frequently (every array with compression/shuffle). Safe to ignore — the V3 spec currently only formally defines the `bytes` (endian) codec, but zarr-python itself uses zstd as the *default* compressor. All practical codecs exist as draft extensions via [zarr-extensions](https://github.com/zarr-developers/zarr-extensions). This is purely a "not yet formally ratified" notice, not a functional issue.

**2. "Consolidated metadata is currently not part in the Zarr format 3 specification"** (UserWarning) — Fires once per `consolidate_metadata()` call (i.e. once per file write). Safe to ignore — consolidated metadata is essential for performant S3/remote access. A formal spec extension is in progress ([zarr-specs PR #309](https://github.com/zarr-developers/zarr-specs/pull/309)).

**3. "The data type ... does not have a Zarr V3 specification"** (FutureWarning) — Fires for dtypes not yet in the V3 spec: fixed-length strings (`<U`), null-terminated bytes (`S`), void/raw bytes (`V`), and structured/compound dtypes. In NWB, this mainly hits compound datasets (electrode tables, trial tables) and fixed-length string arrays. Safe to ignore, though the warning honestly notes the serialization format *could* change in future zarr-python releases. Draft specs exist in [zarr-extensions/data-types](https://github.com/zarr-developers/zarr-extensions/tree/main/data-types).

**4. "Spec '...': Value with data type ... is being converted to data type ..."** (UserWarning) — This is from **hdmf** (`DtypeConversionWarning`), not zarr. It fires when data dtype doesn't exactly match the NWB spec (e.g. Python `int` → `int32`). Not new to this PR — also occurs with the HDF5 backend. Generally safe but could be addressed upstream in hdmf if desired.

In summary: warnings 1-3 all reflect that the zarr V3 spec is intentionally minimal while zarr-python v3 supports a broader feature set as extensions. All features are fully functional and widely used. They occur on every file write but represent spec formalization status, not data correctness issues. I think suppressing them in the gallery tests is the right approach so the tutorials render cleanly — users don't need to see these.

oruebel: It sounds like if a users reads/writes a file this would then result in a lot of warnings from Zarr python. I'm wondering if provide some option (e.g., helper function or option on ZarrIO) to make it easy for user to configure filtering warnings would be useful?

bendichter: maybe, but on the other hand that's really more of a Zarr issue.

oruebel:
> I know AIND is one of the main groups pushing for Zarr v3, so I think from their perspective they would rather us push forward with this

At the very least, I think it would be useful to add some discussion in the docs on: 

- Zarr v2/v3 incompatibility to clarify why this is an issue
- Mention previous version of hdmf-zarr to read V2 files (and when it will stop support)
- It looks like there is a CLI tool for migrating V2 to V3. It would be useful to test whether migration with that tool works https://zarr.readthedocs.io/en/stable/user-guide/cli/ . If that works, then I think that would a good option, because it would at least provide a useful path without introducing a lot of extra burden on hdmf-zarr

src/hdmf_zarr/utils.py
```diff
@@ -476,17 +480,19 @@ def __init__(self, **kwargs):
...
                 if not compressor:
-                    self.__iosettings["compressor"] = None
+                    self.__iosettings["compressors"] = None
                 # To use default settings simply do not specify any compressor settings
                 else:
                     pass
-            # use the user-specified compressor
+            # use the user-specified compressor(s)
+            elif isinstance(compressor, list):
+                self.__iosettings["compressors"] = compressor
```
oruebel:
```suggestion
```

This condition check seems unnecessary since this `elif` and the `else` condition do the same thing

src/hdmf_zarr/utils.py
```diff
@@ -520,14 +526,19 @@ def from_h5py_dataset(h5dataset, **kwargs):
...
-        filters = ZarrDataIO.hdf5_to_zarr_filters(h5dataset)
+        all_codecs = ZarrDataIO.hdf5_to_zarr_filters(h5dataset)
+        # In zarr v3, separate compressors (BytesBytesCodec) from filters (ArrayArrayCodec)
+        compressors = [c for c in all_codecs if isinstance(c, BytesBytesCodec)]
+        filters = [c for c in all_codecs if isinstance(c, ArrayArrayCodec)]
         fillval = h5dataset.fillvalue if "fillvalue" not in kwargs else kwargs.pop("fillvalue")
         if isinstance(fillval, bytes):  # bytes are not JSON serializable so use string instead
             fillval = fillval.decode("utf-8")
         chunks = h5dataset.chunks if "chunks" not in kwargs else kwargs.pop("chunks")
+        compressor = compressors[0] if len(compressors) == 1 else (compressors if compressors else None)
```
oruebel: I think something a little more explicit would be easier to read for this expression:

```suggestion
        if len(compressors) == 1:
            compressor = compressors[0]
        elif compressors:  # len(compressors) > 1
            compressor = compressors
        else:   # len(compressors) == 0
            compressor = None
```

src/hdmf_zarr/utils.py
```diff
@@ -520,14 +526,19 @@ def from_h5py_dataset(h5dataset, **kwargs):
...
         re = ZarrDataIO(
             data=h5dataset,
-            filters=filters,
+            compressor=compressor if compressor else None,
```
oruebel: I think is if/else assignment is redundant. Above the `compressor` varibale is being set either to `None`, `list`, or a single filter object, as such, I think the condition would only ever evaluate to `False`  when `compressor` is None

```suggestion
            compressor=compressor
```

src/hdmf_zarr/zarr_utils.py
oruebel: I'm still making my way through the other changes, so I'll look at the specific change when I get to `backend.py`. One thing to consider is if we want to have minimum length here. In use-cases involving appending rows to data tables, the string length determined form the initial data may be too short so choosing a reasonable minimum length (e.g, 512 or 1024 characters) would help avoid this issue in hopefully most cases. Also, we should mention the use of fixed-length strings the docs; I think the https://hdmf-zarr.readthedocs.io/en/stable/storage.html#dtype-mappings would be a good place to do that since this is specific to references and compound data types.

tests/unit/base_tests_zarrio.py
```diff
@@ -894,8 +884,9 @@ def test_write_structured_array_table(self):
         tempIO.open()
         tempIO.write_dataset(tempIO._file, DatasetBuilder("test_dataset", data, attributes={}, dtype=dt))
         dset = tempIO._file["test_dataset"]
-        self.assertEqual(dset["a"].tolist(), data["a"].tolist())
-        self.assertEqual(dset["b"].tolist(), data["b"].tolist())
+        # In zarr v3, field-name indexing on Array is not supported; read all then index
```
oruebel, line 887: May be something to add to the https://hdmf-zarr.readthedocs.io/en/stable/overview.html#known-limitations

src/hdmf_zarr/zarr_utils.py
oruebel: Another good place to mention this is in the list of known limitations here https://hdmf-zarr.readthedocs.io/en/stable/overview.html#known-limitations

src/hdmf_zarr/backend.py
```diff
@@ -33,6 +31,24 @@
...
+# zarr v3 Array scalar indexing returns 0-d ndarrays instead of numpy scalars;
+# patch to match zarr v2 / numpy behavior expected by hdmf type checks
+_zarr_array_original_getitem = Array.__getitem__
+
+
+def _zarr_array_getitem_scalar_fix(self, key):
+    result = _zarr_array_original_getitem(self, key)
+    if isinstance(result, np.ndarray) and result.ndim == 0:
+        return result[()]
+    return result
```
oruebel, line 52: These look like changes that would be good to propose directly to Zarr. 

It would be great if we could avoid monkey-patching Zarr as this may have unintended side-effects in other libraries that use Zarr. 

I'm not sure how easy/hard it would be to avoid this via changes here and/or in HDMF?

src/hdmf_zarr/backend.py
```diff
@@ -50,7 +66,7 @@
 Default name of the group where specifications should be cached
 """
 
-SUPPORTED_ZARR_STORES = (DirectoryStore, TempStore, NestedDirectoryStore)
+SUPPORTED_ZARR_STORES = (LocalStore,)
```
```suggestion
SUPPORTED_ZARR_STORES = (LocalStore, ) if not FSSPECSTORE_AVAILABLE else (LocalStore, FsspecStore)

```

I don't think `LocalStore` supports reading from S3. I.e. to support remote data read I believe we also need to allow `FsspecStore`

src/hdmf_zarr/backend.py
```diff
@@ -1,20 +1,18 @@
...
 import logging
 
 # Zarr imports
 import zarr
-from zarr.hierarchy import Group
-from zarr.core import Array
-from zarr.storage import DirectoryStore, TempStore, NestedDirectoryStore, ConsolidatedMetadataStore
-import numcodecs
+from zarr import Group, Array
+from zarr.storage import LocalStore
```
oruebel, line 15:
```suggestion
from zarr.storage import LocalStore
try:
    from zarr.storage import FsspecStore
    FSSPECSTORE_AVAILABLE = True
except ImportError:
    FSSPECSTORE_AVAILABLE = False
```

src/hdmf_zarr/backend.py
```diff
@@ -145,49 +152,40 @@ def generate_dataset_html(dataset):
     )
     def __init__(self, **kwargs):
         self.logger = logging.getLogger("%s.%s" % (self.__class__.__module__, self.__class__.__qualname__))
-        path, manager, mode, synchronizer, object_codec_class, storage_options, force_overwrite = popargs(
+        path, manager, mode, storage_options, force_overwrite = popargs(
```
oruebel, line 166: `FsspecStore` is a read-only store, we should check in `__init__` and `if mode != `r` we should either 1) raise an error or 2) change the mode to `r` and warn.

src/hdmf_zarr/backend.py
```diff
@@ -263,12 +249,13 @@ def close(self):
...
-            return True
-        else:
-            return False
+        try:
+            from zarr.storage import FsspecStore
+            if isinstance(self.__file.store, FsspecStore):
+                return True
+        except ImportError:
+            pass
+        return False
```
```suggestion
        if FSSPECSTORE_AVAILABLE and isinstance(self.__file.store, FsspecStore):
            return True
        else:
            return False
```

src/hdmf_zarr/backend.py
```diff
@@ -310,8 +297,12 @@ def load_namespaces(cls, namespace_catalog, path, file, storage_options, namespa
             raise ValueError("Only one of 'path' and 'file' must be provided.")
 
         if path is not None:
-            # TODO: how to use storage_options here?
-            f = zarr.open(path, mode="r", storage_options=storage_options)
+            if storage_options is not None:
```
oruebel: Is `if storage_options is not None:` the correct condition to use here? It looks like the goal is to distinguish between a local and a remote path, i.e.,

1)  shouldn't this `if` condition look at the `path` instead
2) this implies that `storage_options` are required when using the `FsspecStore`. If that is the case then we should mention that in the doc for the `storage_options` parameter. If they are not required, then we should update the `if` condition to not make them required.

src/hdmf_zarr/backend.py
```diff
@@ -642,17 +633,23 @@ def write_builder(self, **kwargs):
...
-        else:
-            fpath = store.path
-
-        return fpath
 
-    def __open_file_consolidated(self, store, mode, synchronizer=None, storage_options=None):
+        For local stores, this returns the resolved absolute filesystem path.
+        For remote stores, this returns the string representation.
+        """
+        from zarr.storage import LocalStore
```
```suggestion
```

`LocalStore` is already a required import at the top of the module

src/hdmf_zarr/backend.py
```diff
@@ -642,17 +633,23 @@ def write_builder(self, **kwargs):
...
+        if isinstance(store, LocalStore):
+            return str(store.root.resolve())
+        return str(store)
+
+    def __open_file(self, store, mode, storage_options=None):
+        """Open a zarr file without consolidated metadata."""
+        if storage_options is not None:
```
oruebel: Same comment as above, I'm not sure that `if storage_options` is the right approach to check for a remote file. I would suggest adding a small helper function along the lines of:

```

def __is_remote(self, store) -> bool:
    """
    Determine if a given store is remote.

    Args:
    store: A store object or path (str or Path).

    Returns:
    bool: True if the store is remote, False otherwise.
    """

    # If store is a string or Path, check if it's a local file path
    if isinstance(store, (str, Path)):
        return not store.startswith('file://') and not Path(store).is_absolute() and not store.startswith('/')

    # If store is a Zarr LocalStore, it's not remote
    if isinstance(store, zarr.storage.LocalStore):
        return False
    else:
        return True
```

src/hdmf_zarr/backend.py
```diff
@@ -662,29 +659,31 @@ def __open_file_consolidated(self, store, mode, synchronizer=None, storage_optio
...
         # Create a cache key that includes relevant parameters
-        cache_key = (store_path, mode, str(synchronizer), str(storage_options))
+        cache_key = (store_path, mode, str(storage_options))
 
         # Check if we already have this file cached
         if cache_key in self.__consolidated_cache:
             return self.__consolidated_cache[cache_key]
 
         # Open the file and cache the result
+        if storage_options is not None:
```
oruebel: Same comment as above, use a helper function to determine whether to use FSSpec instead of `if storage_options is not None`

src/hdmf_zarr/backend.py
```diff
@@ -800,7 +799,7 @@ def write_attributes(self, **kwargs):
             # Case 2: References
             elif isinstance(value, (Builder, ReferenceBuilder)):
                 refs = self._create_ref(value, ref_link_source=self.path)
-                tmp = {"zarr_dtype": "object", "value": refs}
+                tmp = {"zarr_dtype": "object", "value": dict(refs)}
```
oruebel, line 816: Just curious, why is `dict(refs)` needed?  `refs = self._create_ref(...` returns a `ZarrReference(dict)`, which is already a dict, so why do we need to convert again to a dict?

src/hdmf_zarr/backend.py
oruebel: It looks like the expected input types for the `zarr_ref` parameter have changed, so we should update the docstring accordingly

```suggestion
        :param zarr_ref: Either: 1) Dict with `source` and `path` keys, or 2) a `ZarrReference` object, or 3) a JSON string encoding the dict, or 4) a scalar np.ndarray of StringDType
```

src/hdmf_zarr/backend.py
```diff
@@ -871,6 +870,12 @@ def resolve_ref(self, zarr_ref):
         :return: 1) name of the target object
                  2) the target zarr object within the target file
         """
+        # In zarr v3, references may be stored as JSON strings or numpy StringDType scalars
+        if isinstance(zarr_ref, np.ndarray):
+            zarr_ref = str(zarr_ref)
```
oruebel: I think we need to load the scalar value for the numpy array. We may also want to check that this is indeed a scalar
```suggestion
        if isinstance(zarr_ref, np.ndarray):
            if zarr_ref.ndim == 0
                zarr_ref = str(zarr_ref[()])
            else:
                raise ValueError("ZarrIO.resolve_ref expects a single reference not a multi-valued array")
```

src/hdmf_zarr/backend.py
```diff
@@ -952,12 +957,10 @@ def _create_ref(self, ref_object, ref_link_source=None):
                 builder.source if (builder.source is not None and os.path.isdir(builder.source)) else self.source
             )
 
-        if not isinstance(ref_link_source, str):
-            # self.path is sometimes given as the ref_link_source. It can
-            # be either a (str, Path, *SUPPORTED_ZARR_STORES). That being said,
-            # when self.path is a Path, it is converted to a str in __init__.
-            # We only have to deal with *SUPPORTED_ZARR_STORES and strings.
```
oruebel, line 959: I believe this comment was removed by accident and should remain

src/hdmf_zarr/backend.py
```diff
@@ -997,7 +1002,7 @@ def __add_link__(self, parent, target_source, target_path, link_name):
             parent.attrs["zarr_link"] = []
         zarr_link = list(parent.attrs["zarr_link"])
         if not isinstance(target_source, str):  # a store
-            target_source = target_source.path
+            target_source = str(target_source)
```
oruebel: Does a Zarr store correctly convert to a path with just `str`?

src/hdmf_zarr/backend.py
```diff
@@ -1020,15 +1025,6 @@ def write_link(self, **kwargs):
...
             if target_builder.source in (group_filename, export_source):
-                # Case 1:
-                # target_builder.source == export_source
-                # This means we have a SoftLink for a group and so we want the exported link to
-                # also point "inwards" in the file being created.
-                #################################
-                # Case 2:
-                # target_builder.source == group_filename
-                # This is still a SoftLink; however, it is from adding a link to a group after FileA
-                # has been read and we are exporting that to FileB. We still want the link to be "inwards".
```
oruebel, line 1031: I think this comment is still valid and should probably remain

src/hdmf_zarr/backend.py
```diff
@@ -1049,16 +1045,7 @@ def write_link(self, **kwargs):
...
-        Setup a dataset for writing to one-chunk-at-a-time based on the given DataChunkIterator. This
-        is a helper function for write_dataset()
-        :param parent: The parent object to which the dataset should be added
-        :type parent: Zarr Group or File
-        :param name: The name of the dataset
-        :type name: str
-        :param data: The data to be written.
-        :type data: AbstractDataChunkIterator
-        :param options: Dict with options for creating a dataset. available options are 'dtype' and 'io_settings'
-        :type options: dict
```
oruebel, line 1061: I believe this docstring may have been removed by accident

src/hdmf_zarr/backend.py
```diff
@@ -1067,25 +1054,66 @@ def __setup_chunked_dataset__(cls, parent, name, data, options=None):
         # Define the chunking options if the user has not set them explicitly. We need chunking for the iterative write.
         if "chunks" not in io_settings:
             recommended_chunks = data.recommended_chunk_shape()
-            io_settings["chunks"] = True if recommended_chunks is None else recommended_chunks
+            # In zarr v3, chunks=True is not valid; omit to use zarr defaults
+            if recommended_chunks is not None:
+                io_settings["chunks"] = recommended_chunks
```
```suggestion
            if recommended_chunks is not None:
                io_settings["chunks"] = recommended_chunks
            else:
                # omit chunking option to use zarr defaults. Similar to chunks=True in HDF5. 
                pass
```

src/hdmf_zarr/backend.py
```diff
@@ -1067,25 +1054,66 @@ def __setup_chunked_dataset__(cls, parent, name, data, options=None):
...
+            if (options is not None) and ("dtype" in options) and options["dtype"] is not None:
                 io_settings["dtype"] = options["dtype"]
             else:
                 io_settings["dtype"] = data.dtype
             if isinstance(io_settings["dtype"], str):
                 # map to real dtype if we were given a string
                 io_settings["dtype"] = cls.__dtypes.get(io_settings["dtype"])
+            # zarr v3 requires dtype to be specified explicitly
+            if io_settings["dtype"] is None:
+                io_settings["dtype"] = np.float64
```
oruebel: Hardcoding the dtype here to `np.float64` I believe would falsely change the dtype for write and would likely cause an error. If the original docstring for this function is still correct, then the `data` input is an `AbstractDataChunkIterator` so we should get the actual dtype from that instead. 

```suggestion
            # zarr v3 requires dtype to be specified explicitly
            if io_settings["dtype"] is None:
                io_settings["dtype"] = data.dtype
```

oruebel: Actually, I think this check could be removed entirely. Above the `io_setting["dtype"]` is being set via:

```
           if (options is not None) and ("dtype" in options) and options["dtype"] is not None:
                io_settings["dtype"] = options["dtype"]
            else:
                io_settings["dtype"] = data.dtype
``` 

If  io_settings["dtype"] would be None afterwards, I think that would be an error because `data.dtype` should never be None and the `options["dtype"]` is only being used if it is not None

test_gallery.py
bendichter: These warnings come from zarr-python itself:

1. **Consolidated metadata warning** — occurs on every read with open_consolidated(). Expected since consolidated metadata isn't part of the zarr v3 spec yet.
2. **Unstable dtype warning** — occurs when using compound dtypes with FixedLengthUTF32 fields. Only affects compound dtypes.
3. **Numcodecs codecs warning** — occurs with HDF5-to-Zarr conversion using numcodecs wrappers.

These are informational warnings about unstable/unspecified zarr v3 features. They're safe to ignore — operations complete correctly. They'll likely go away as zarr v3 matures.

bendichter: Good idea. I think adding a warning filter helper or option would be a nice enhancement for a follow-up PR. For now, users can use standard Python `warnings.filterwarnings()` to suppress these.

src/hdmf_zarr/utils.py
bendichter: Fixed in 8a0195d — removed the redundant `elif isinstance(compressor, list)` branch.

src/hdmf_zarr/utils.py
bendichter: Fixed in 8a0195d — changed to explicit if/elif/else as suggested.

src/hdmf_zarr/utils.py
bendichter: Fixed in 8a0195d — the compressor variable is now set cleanly in the if/elif/else block above, so the conditional on the next line is no longer needed.

src/hdmf_zarr/zarr_utils.py
bendichter: Great point about appending. Fixed in 8a0195d — added `COMPOUND_DTYPE_MIN_STRING_LENGTH = 512` module constant, used as the minimum for all string/reference fields in compound dtypes. Also documented in storage.rst dtype mappings section.

tests/unit/base_tests_zarrio.py
bendichter: Added to known limitations in 8a0195d — documented both the FixedLengthUTF32 compound dtype limitation and the zarr v3-only write format.

src/hdmf_zarr/zarr_utils.py
bendichter: Done — added to known limitations and documented in storage.rst. See 8a0195d.

src/hdmf_zarr/backend.py
bendichter: Agreed, these should be proposed upstream to zarr-python. The two patches are:
1. `Array.__len__` — zarr v3 Array doesn't implement `__len__`, but HDMF and many array-like interfaces expect it.
2. `Array.__getitem__` scalar fix — zarr v3 returns 0-d ndarrays instead of scalars, which breaks HDMF type checks.

I'll file issues/PRs upstream. In the meantime, the monkey-patches are the least invasive way to maintain compatibility.

src/hdmf_zarr/backend.py
bendichter: Done in 8a0195d — FsspecStore import now uses try/except with `FSSPECSTORE_AVAILABLE` flag at module level.

src/hdmf_zarr/backend.py
bendichter: Fixed in 8a0195d — `SUPPORTED_ZARR_STORES` now includes `FsspecStore` when available.

src/hdmf_zarr/backend.py
bendichter: Fixed in 8a0195d — `__init__` now raises a `ValueError` if `storage_options` is provided with a mode other than `'r'`.

src/hdmf_zarr/backend.py
bendichter: Fixed in 8a0195d — `is_remote()` now checks `FSSPECSTORE_AVAILABLE` before using `isinstance`.

src/hdmf_zarr/backend.py
bendichter: Fixed in 8a0195d — added a `__resolve_store` static helper that centralizes the FsspecStore creation logic. All `if storage_options is not None` checks now go through this helper. The helper also checks `FSSPECSTORE_AVAILABLE` and raises a clear `ImportError` if fsspec is needed but not installed.

src/hdmf_zarr/backend.py
bendichter: Removed in 8a0195d.

src/hdmf_zarr/backend.py
bendichter: Fixed in 8a0195d — added `__resolve_store` static helper that all three callsites now use.

src/hdmf_zarr/backend.py
bendichter: Fixed — all three callsites now use the `__resolve_store` helper. See 8a0195d.

src/hdmf_zarr/backend.py
bendichter: Good question. It turns out `dict(refs)` IS needed here. `ZarrReference` is a dict subclass but its `__init__` uses `docval` with named parameters. When zarr serializes attributes to JSON, it uses `dataclasses._asdict_inner` which calls `type(obj)(generator)` to reconstruct dicts — this fails on `ZarrReference` because its `__init__` expects specific named args, not a generator. Converting to plain `dict` avoids this. Added a comment explaining why.

src/hdmf_zarr/backend.py
bendichter: Updated in 8a0195d — docstring now documents all accepted types: dict, ZarrReference, JSON string, and scalar np.ndarray of StringDType.

src/hdmf_zarr/backend.py
bendichter: Fixed in 8a0195d — now uses `zarr_ref[()]` to extract the scalar value and raises `ValueError` if the ndarray is not scalar (ndim != 0).

bendichter:
Addressed all 20 inline review comments in 8a0195d. Summary of changes:

**Code cleanup:**
- Removed redundant `elif` and simplified compressor assignment in `utils.py`
- Added `__resolve_store` helper to consolidate all `storage_options` handling
- Moved `FsspecStore` import to top-level with `FSSPECSTORE_AVAILABLE` flag
- Added `FsspecStore` to `SUPPORTED_ZARR_STORES`, read-only mode enforcement
- Updated `resolve_ref` docstring, added scalar ndarray validation
- Explained why `dict(refs)` is needed (ZarrReference + dataclasses serialization)

**Compound dtypes:**
- Added `COMPOUND_DTYPE_MIN_STRING_LENGTH = 512` for fixed-length string fields

**Documentation:**
- Updated known limitations for zarr v3 and compound dtype strings
- Documented fixed-length string usage in storage.rst dtype mappings
- Added zarr CLI migration tool link as recommended

Re: zarr CLI tool — I haven't tested it with NWB-Zarr files yet. The main concern is whether it handles `object_codec` datasets. I suspect it won't (since that's a zarr v2 concept), but it would migrate everything else. Happy to test if you think that's worth doing in this PR vs. a follow-up.

src/hdmf_zarr/backend.py
bendichter: Restored in 5d83ba7.

src/hdmf_zarr/backend.py
bendichter: Good catch — `str(LocalStore(...))` returns a URI (`file:///...`), not a filesystem path. Fixed in 7f2f72d to use `__get_store_path()` which properly resolves to a filesystem path for LocalStore.

src/hdmf_zarr/backend.py
bendichter: Restored in 089ff3a.

src/hdmf_zarr/backend.py
bendichter: Restored in 089ff3a.

src/hdmf_zarr/backend.py
bendichter: Applied in efa03f3.

src/hdmf_zarr/backend.py
bendichter: Agreed and taken a step further per your follow-up comment — removed the fallback entirely since `data.dtype` should never be None and `options["dtype"]` is only used when not None. See ff4e80d.

bendichter: Good point — removed the fallback entirely in ff4e80d.

tests/unit/test_fsspec_streaming.py
```diff
@@ -2,11 +2,10 @@
 from hdmf_zarr import NWBZarrIO
 from .helpers.utils import check_s3fs_ffspec_installed
 
-import zarr
-
 HAVE_FSSPEC = check_s3fs_ffspec_installed()
 
 
+@unittest.skip("S3 test files are in zarr v2 format with object_codec which is not supported by zarr v3")
```
oruebel, line 8: @bendichter we should add a Zarr V3 test dataset to DANDI to test streaming read. Maybe as part of https://dandiarchive.org/dandiset/001083. We should at least file an issue for this. 

In addition, as part of this PR, is there a way to check that a file is V2 when we try to access it (probably in `ZarrIO.open` or `ZarrIO.__init__`) to check that the file is V2 and if it is V2 raise an error. If reading V2 is not supported then we should raise an error as early as possible to avoid confusion.

src/hdmf_zarr/backend.py
h-mayorquin: I filed both issues upstream:
- `__len__` removal: https://github.com/zarr-developers/zarr-python/issues/3740
- `__getitem__` scalar indexing: https://github.com/zarr-developers/zarr-python/issues/3741

I think both of these come from zarr v3 aligning with the [Python array API standard](https://data-apis.org/array-api/latest/), which does not include `__len__` on array objects and requires scalar indexing to return 0-d arrays instead of scalars. 

I think the monkey-patching is too wide of a fix and probably dangerous because it will leak as side effects to any other zarr object used in the same process. I think we can handle this in a better way with two changes in hdmf:

1. https://github.com/hdmf-dev/hdmf/pull/1414 replaces the collection detection through duck typing of `hasattr(data, "__len__")` with checks on `ndim`, which is a more general attribute.
2. https://github.com/hdmf-dev/hdmf/pull/1415 adds guards for 0-d ndarrays in the type inference code, removing the need for the `__getitem__` monkey-patch.

Together, these would make hdmf work correctly with zarr v3 (and any other array API compliant library) without needing to modify zarr's behavior globally. Plus, they will bring the hdmf codebase one step closer to working with any array library that supports the array API standard.

bendichter:
Zarr 3.1.3 doesn't support Python 3.10, so the required checks will need to be bumped to 3.11 for this PR

yarikoptic:
note, that it seems that hdmf-zarr restriction of numcodes seems to forbid dandi-cli to come to the bright future of python 3.14:

- https://github.com/dandi/dandi-cli/pull/1838

Is there a time line for this PR or should we look into alternative approaches on how to unblock us? (e.g. proposing hdmf-zarr to become an optional dependency, or alike)

rly:
> Is there a time line for this PR or should we look into alternative approaches on how to unblock us? (e.g. proposing hdmf-zarr to become an optional dependency, or alike)

We discussed this on Monday. https://github.com/NeurodataWithoutBorders/nwbinspector/issues/698 should unblock dandi

alejoe91:
This PR is tested against also in SpikeInterface (all green!): https://github.com/SpikeInterface/spikeinterface/pull/4260

Also, here is a WIP PR in neuroconv to configure a Zarr v3 backend with new codecs and sharding: https://github.com/catalystneuro/neuroconv/pull/1749

src/hdmf_zarr/backend.py
```diff
@@ -1739,6 +1847,10 @@ def __read_dataset(self, zarr_obj, name):
         elif self.__is_ref(dtype):
             # Array of references
             data = BuilderZarrReferenceDataset(data, self)
+        # Eagerly load StringDType arrays so other backends (e.g. HDF5IO) can handle them.
+        # Must be after reference checks since references are also stored as StringDType.
+        elif isinstance(zarr_obj.dtype, np.dtypes.StringDType) and dtype != "scalar":
+            data = list(zarr_obj[:])
```
bendichter: Reading any non-scalar string dataset now does `data = list(zarr_obj[:])`. For a large 1-D text column (e.g. millions of entries) this pulls the whole dataset into a Python list on open, defeating zarr's chunked/lazy access and risking OOM. Worse, for a 2-D string dataset `list(zarr_obj[:])` yields a list of row ndarrays, so a consumer doing `builder.data[i, j]` raises `TypeError: list indices must be integers or slices, not tuple`, and `.shape`/`.dtype` are gone. Consider a lazy decoding wrapper (like the reference-dataset wrappers just above) instead of full materialization.

src/hdmf_zarr/backend.py
```diff
@@ -148,49 +167,45 @@ def generate_dataset_html(dataset):
...
-            else:
-                self.__synchronizer = None
-        else:
-            self.__synchronizer = synchronizer
-        self.__mode = mode
         self.__force_overwrite = force_overwrite
         if isinstance(path, Path):
             path = str(path)
+        # Convert local paths to absolute for consistent path resolution
+        if isinstance(path, str) and not path.startswith(("s3://", "http://", "https://")):
```
bendichter: Any other fsspec protocol (`gcs://`, `gs://`, `abfs://`, `az://`) fails the check and gets rewritten to `/cwd/gcs:/bucket/file.zarr`, so the subsequent `FsspecStore.from_url` receives a corrupted URL and the open fails or points at the wrong location. A GCS/Azure user passing `storage_options` cannot open their file.

src/hdmf_zarr/backend.py
```diff
@@ -1085,25 +1112,65 @@ def __setup_chunked_dataset__(cls, parent, name, data, options=None):
...
+        # Copy compressors/codecs if available (zarr v3 uses 'compressors' plural)
+        if hasattr(source, 'compressors') and source.compressors:
+            kwargs['compressors'] = list(source.compressors)
+        if hasattr(source, 'filters') and source.filters:
+            kwargs['filters'] = list(source.filters)
+        if hasattr(source, 'fill_value'):
+            kwargs['fill_value'] = source.fill_value
+
+        dest = dest_group.create_array(**kwargs)
+        dest[:] = source[:]
```
bendichter: The replaced `zarr.copy()` streamed chunk-by-chunk. During export of a multi-GB dataset (e.g. a large ElectricalSeries), this now decompresses the whole array into one in-memory ndarray before writing, which can OOM. Iterate over chunk regions (`for sl in ...: dest[sl] = source[sl]`) to bound memory to a single chunk. (zarr v3's `zarr.from_array(...)` also does this and would copy codecs/fill_value for you.)

src/hdmf_zarr/backend.py
bendichter: Good catch — fixed in b23a49a.

Replaced the eager `data = list(zarr_obj[:])` with a lazy `ZarrStringDataset` wrapper (in `zarr_utils.py`, mirroring the reference-dataset wrappers just above), so string datasets are no longer materialized on open:

- **Lazy** — wrapping reads nothing; elements are decoded to native `str` only on access, preserving zarr's chunked access and avoiding the OOM risk on large text columns.
- **2-D correctness** — `data[i, j]` now returns the decoded scalar (previously `TypeError: list indices must be integers or slices, not tuple`), and `.shape`/`.dtype`/`len()` are preserved.
- **Export** — reports an object dtype so HDMF infers `utf8` on re-build; Zarr→HDF5 and Zarr→Zarr round-trips verified.

(Root cause of why the eager cast was there: for generic `dtype=None` specs like `VectorData`, HDMF's `__check_edgecases` doesn't recognize `np.dtypes.StringDType` and falls through to `value.dtype.type` = python `str`, which HDF5IO can't write. The wrapper's object dtype steers inference to `utf8` instead.)

Added `TestZarrStringDataset` covering laziness, shape/dtype/len preservation, scalar/slice decoding, and 2-D `[i, j]` indexing.

src/hdmf_zarr/backend.py
bendichter: Fixed in ed79899. The check now detects any protocol URL generically via `"://" not in path` instead of an s3/http allowlist, so `gcs://`, `gs://`, `abfs://`, `az://`, and chained fsspec protocols (e.g. `simplecache::s3://`) are left untouched while local/relative paths are still made absolute. Added `TestPathNormalization` covering all these protocols.

src/hdmf_zarr/backend.py
bendichter: Fixed in ed79899. `_copy_array` now copies one chunk region at a time (`itertools.product` over chunk-aligned offsets → `dest[region] = source[region]`) so peak memory is bounded to a single chunk instead of decompressing the whole array into one ndarray. I kept the explicit copy rather than `zarr.from_array` since this method already handles the object→`StringDType` conversion and copies compressors/filters/fill_value/attrs; the chunk loop reuses `dest.chunks` (== `source.chunks`). Added `TestCopyArray` covering multi-chunk numeric, object→StringDType, and scalar (0-d) copies.

src/hdmf_zarr/backend.py
bendichter: Correction / follow-up (b3dbd30): the lazy `ZarrStringDataset` wrapper I described above broke export of **typed** text columns (spec dtype `text`, e.g. `electrodes.location`) and failed the `plot_convert_nwb_hdf5` gallery job. Root cause: HDMF's `convert_dtype` only recognizes `zarr.Array` / `ndarray` / `StrDataset` / `list` / `DataChunkIterator` for string specs, so a custom wrapper falls through to `_unicode(wrapper)` and raises `Expected unicode or ascii string`. A raw zarr StringDType array doesn't work either: its dtype kind is `"T"`, which HDMF's generic (`dtype=None`) inference doesn't recognize, so it infers python `str` and HDF5IO can't create the dataset.

I've replaced the wrapper with decoding to a **numpy object array** of native `str` on read. That is recognized by HDMF for both typed and generic specs, preserves `shape`/`dtype`, and supports N-D `data[i, j]` indexing (which `list(zarr_obj[:])` did not). It is still materialized on open — genuinely lazy access isn't achievable without upstream HDMF support for `StringDType` (or wrapping as a `DataChunkIterator`, which is a much larger change and unsuitable as the read representation). Left a code comment to that effect; happy to file an HDMF issue for the upstream fix if you'd like.

ehennestad:
I hit a compound-dtype bug while testing this branch and filed it as #364.

It is **not a regression from this PR** — `ZarrIO.__dtypes` is missing a bare `uint` key on `dev`
too, so any compound field declared `dtype: uint` is silently written as `float64`. In
hdmf-common-schema that is all five HERD index fields, which breaks the HERD round-trip. On this
branch it surfaces as:

```
TypeError: incorrect type for 'files_idx' (got 'float64', expected 'int or integer')
```

Flagging it here only because this PR reworks the compound-dataset write path, so it may be a
convenient moment to fix it. The change itself is small (add `"uint": np.uint32`, matching
`HDF5IO`), and it would be caught by a test asserting that a written compound dtype matches the
spec rather than only that the values round-trip.

Otherwise the branch has been working well for me — I ported 14 PyNWB gallery tutorials to Zarr v3
against it and everything else passes.

src/hdmf_zarr/backend.py
oruebel: https://github.com/hdmf-dev/hdmf-zarr/pull/366/commits/5d6ed2d7ba849f811b0384502f20575978d3803d adapts the proposed approach for ZarrStringArray by using a new `HDMFZarrArray` wrapper class that inherits from Zarr array so that strings are read lazily.

docs/gallery/plot_zarr_io.py
```diff
@@ -157,29 +156,29 @@
...
 #
 
-from zarr.storage import NestedDirectoryStore
+from zarr.storage import LocalStore
 from hdmf.common import CSRMatrix
 
-zarr_nsd_dir = "example_nested_store.zarr"
-store = NestedDirectoryStore(zarr_dir)
+zarr_ls_dir = "example_local_store.zarr"
+store = LocalStore(zarr_ls_dir)
```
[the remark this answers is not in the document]
oruebel: Fixed in [0f48a0d](https://github.com/hdmf-dev/hdmf-zarr/pull/325/commits/0f48a0ddb690fdfbdb1acd3def946874ac702f70)

src/hdmf_zarr/utils.py
```diff
@@ -520,14 +530,24 @@ def from_h5py_dataset(h5dataset, **kwargs):
...
+            compressor = compressors[0]
+        elif len(compressors) > 1:
+            compressor = compressors
+        else:
+            compressor = None
         re = ZarrDataIO(
             data=h5dataset,
-            filters=filters,
+            compressor=compressor,
+            filters=filters if filters else None,
```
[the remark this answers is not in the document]
oruebel: Fixed in [eb003c8](https://github.com/hdmf-dev/hdmf-zarr/pull/366/commits/eb003c863ff720d8549ad6693ed64d9aa2afbcc2)

src/hdmf_zarr/nwb.py
```diff
@@ -34,26 +67,11 @@ class NWBZarrIO(ZarrIO):
...
-        else:
-            if manager is not None and extensions is not None:
-                raise ValueError("'manager' and 'extensions' cannot be specified together")
-            elif extensions is not None:
-                manager = get_manager(extensions=extensions)
-            elif manager is None:
-                manager = get_manager()
-        super().__init__(path, manager=manager, mode=mode, synchronizer=synchronizer, storage_options=storage_options)
+        manager = _build_nwb_manager(type(self), path, mode, manager, extensions, load_namespaces, storage_options)
+        super().__init__(path, manager=manager, mode=mode, storage_options=storage_options)
```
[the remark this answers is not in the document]
oruebel: For V2 files we would use `NWBZarrV2IO` from `nwb_zarrV2.py` instead. Since this is a major updated and since `synchronizer` is rarely used, it is fine to remove the argument here and implement the V3 logic directly.

src/hdmf_zarr/backend_zarrv2.py
```diff
@@ -0,0 +1,637 @@
...
+            # reshaping to chunk_shape. This assumes object-dtype v2 arrays are 1-D, which
+            # holds for every array this fallback handles (specs, references, vlen columns).
+            # A multi-dimensional object-dtype array split across chunks would misalign
+            # against the N-D slices in _decode_v2_dataset; add reshaping here if such an
+            # array is ever encountered.
+            for filt in reversed(filters):
+                raw = filt.decode(raw)
+            if isinstance(raw, np.ndarray):
+                return raw
+            return np.array(raw, dtype=object)
```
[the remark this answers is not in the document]
oruebel: Fixed in [8a760bd](https://github.com/hdmf-dev/hdmf-zarr/pull/366/commits/8a760bd0f2c3c8695b50624c716d8fa55b5f3c59)

src/hdmf_zarr/backend_zarrv2.py
```diff
@@ -0,0 +1,637 @@
...
+    Checks the zarr format reported by the opened hierarchy (zarr v2 sets
+    ``zarr_format=2`` in ``.zgroup``; zarr v3 uses ``zarr.json``).  Works for
+    local paths, opened Zarr stores, and remote URLs (``s3://``,
+    ``http(s)://``, ``gs://``).
+    """
+    if isinstance(path, (str, os.PathLike)) and not isinstance(path, LocalStore):
+        path_str = str(path)
+        is_remote = path_str.startswith(("s3://", "http://", "https://", "gs://"))
+        if not is_remote and storage_options is None:
+            return any(os.path.exists(os.path.join(path_str, m)) for m in (".zgroup", ".zarray"))
```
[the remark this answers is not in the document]
oruebel: Fixed in [af78381](https://github.com/hdmf-dev/hdmf-zarr/pull/366/commits/af783814cf4ccdc7629aefa942a8f59d4d9aa615)

src/hdmf_zarr/backend_zarrv2.py
```diff
@@ -0,0 +1,637 @@
...
+            chunk_name = dimension_separator.join("0" for _ in range(ndim))
+            raw = _read_store_bytes(store, f"{dataset_key}/{chunk_name}")
+            if raw is not None:
+                data = ZarrV2IO._decode_v2_chunk(raw, compressor, filters, dtype, chunks, order, is_object)
+                if chunks != shape:
+                    data = data[tuple(slice(0, s) for s in shape)]
+                return data
+            return np.full(shape, fill_value=None if is_object else 0, dtype=object if is_object else dtype)
+
+        result = np.empty(shape, dtype=object if is_object else dtype)
```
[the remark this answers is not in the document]
oruebel: Fixed in [c82b832](https://github.com/hdmf-dev/hdmf-zarr/pull/366/commits/c82b832221202da6c2095479e80d53b173876071)

src/hdmf_zarr/backend_zarrv2.py
```diff
@@ -0,0 +1,637 @@
...
+            import numcodecs
+
+            raw = numcodecs.get_codec(compressor_config).decode(raw)
+
+        filters = zarray_meta.get("filters") or []
+        if filters:
+            import numcodecs
+
+            for filt_config in reversed(filters):
+                raw = numcodecs.get_codec(filt_config).decode(raw)
```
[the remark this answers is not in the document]
oruebel: Fixed in [13c568d](https://github.com/hdmf-dev/hdmf-zarr/pull/366/commits/13c568d3173e7fd80c3113922485eba946ebf7ff)

src/hdmf_zarr/backend.py
oruebel: This is being addressed in #366

oruebel:
@alejoe91 @h-mayorquin @bendichter @rly I started preparing a short news item on the nwb.org website for this release at https://github.com/NeurodataWithoutBorders/neurodatawithoutborders.github.io/pull/180

rly (dismissed):

rly (approved):

alejoe91:
Yaaaaayyyy :rocket:

PR #328, opened 2026-03-02
Fix is_remote() returning False for HTTPS URLs w/ consolidated metadata

### LINKED ISSUE #327: [Bug]: Error in S3 Streaming Tutorial
### What happened?

Likely an issue with reading consolidated metadata.

### Steps to Reproduce

```shell
python /Users/rly/Documents/NWB/hdmf-zarr/docs/gallery/plot_s3_streaming.py
```

### Traceback

```shell
/Users/rly/Documents/NWB/hdmf-zarr/docs/gallery/plot_s3_streaming.py                                   
  Note: Could not access S3 file (network access may be required): nothing found at path ''              
  Note: Could not access S3 file (network access may be required): nothing found at path ''
```

### Operating System

macOS

### Python Version

3.14

### Package Versions

_No response_

rly:
When consolidated metadata is used, `zarr.open_consolidated()` wraps the underlying `FSStore` in a `ConsolidatedMetadataStore`. This caused `is_remote()` to return False for remote HTTPS stores, which led `resolve_ref()` to mangle URLs with `os.path.normpath`, producing `PathNotFoundError` when reading links.

Unwrap `ConsolidatedMetadataStore` before checking the store type, consistent with the pattern already used elsewhere in the codebase.

Fixes #327

## Checklist

### DIFF
--- a/tests/unit/test_fsspec_streaming.py
+++ b/tests/unit/test_fsspec_streaming.py
@@ -44,6 +44,22 @@ class TestFSSpecStreaming(unittest.TestCase):
             self.assertIsInstance(read_io._file.store, zarr.storage.FSStore)
 
     @unittest.skipIf(not HAVE_FSSPEC, "fsspec not installed")
+    def test_is_remote_with_consolidated(self):
+        """Test that is_remote() returns True for remote HTTPS stores with consolidated metadata."""
+        with NWBZarrIO(self.https_s3_path, mode="r") as read_io:
+            read_io.open()
+            self.assertIsInstance(read_io._file.store, zarr.storage.ConsolidatedMetadataStore)
+            self.assertTrue(read_io.is_remote())
+
+    @unittest.skipIf(not HAVE_FSSPEC, "fsspec not installed")
+    def test_is_remote_without_consolidated(self):
+        """Test that is_remote() returns True for remote HTTPS stores without consolidated metadata."""
+        with NWBZarrIO(self.https_s3_path, mode="-r") as read_io:
+            read_io.open()
+            self.assertIsInstance(read_io._file.store, zarr.storage.FSStore)
+            self.assertTrue(read_io.is_remote())
+
+    @unittest.skipIf(not HAVE_FSSPEC, "fsspec not installed")
     def test_fsspec_streaming_via_read_nwb(self):
         """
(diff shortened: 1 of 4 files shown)


h-mayorquin (dismissed):
LGTM, do we need a test?

I think that zarr 3 does not have this weird wrapping pattern.

rly:
> LGTM, do we need a test?

Yes, thanks!

tests/unit/test_fsspec_streaming.py
```diff
@@ -43,6 +43,22 @@ def test_s3_open_with_consolidated_(self):
             read_io.open()
             self.assertIsInstance(read_io._file.store, zarr.storage.FSStore)
 
+    @unittest.skipIf(not HAVE_FSSPEC, "fsspec not installed")
+    def test_is_remote_with_consolidated(self):
+        """Test that is_remote() returns True for remote HTTPS stores with consolidated metadata."""
+        with NWBZarrIO(self.https_s3_path, mode="r") as read_io:
+            read_io.open()
+            self.assertIsInstance(read_io._file.store, zarr.storage.ConsolidatedMetadataStore)
```
h-mayorquin, line 51: I will remove this line as this is an implementation detail that might change with zarr 3. The important thing is that `read_io.is_remote()`

tests/unit/test_fsspec_streaming.py
```diff
@@ -43,6 +43,22 @@ def test_s3_open_with_consolidated_(self):
...
+            self.assertTrue(read_io.is_remote())
+
+    @unittest.skipIf(not HAVE_FSSPEC, "fsspec not installed")
+    def test_is_remote_without_consolidated(self):
+        """Test that is_remote() returns True for remote HTTPS stores without consolidated metadata."""
+        with NWBZarrIO(self.https_s3_path, mode="-r") as read_io:
+            read_io.open()
+            self.assertIsInstance(read_io._file.store, zarr.storage.FSStore)
```
h-mayorquin, line 59: same here.

h-mayorquin (approved):
I think we should focus the tests on the important behavior that is_remote returns True even for consolidate metadata, we should avoid testing the internal behavior of how the store is represented. 

Otherwise looks good to me.

PR #330, opened 2026-03-02
Update s3 streaming tutorial with faster file

rly:
## Motivation

S3 streaming tutorial was pointing to a zarr file that took a while to load and had extensions. This updated tutorial takes about 10 seconds to run instead of 140 seconds.

Requires #328 to actually run

## How to test the behavior?
```
```

## Checklist


rly:
HDMF-dev CI failures will be addressed in a separate PR.

oruebel (approved):

Issue #335, opened 2026-03-20
Unified Zarr v3 convention: align hdmf-zarr and LINDI storage formats


bendichter:
## Context

**Problem:** hdmf-zarr and LINDI both represent data types that Zarr doesn't natively support (links, references, compound types, scalars) but use incompatible conventions. This prevents using LINDI's Zarr output directly with hdmf-zarr's `NWBZarrIO`, forcing LINDI to maintain an h5py shim layer that causes edge-case bugs.

**Opportunity:** Both libraries are breaking backwards compatibility for Zarr v2 → v3 migration. This is the ideal time to align on a single standard.
- hdmf-zarr (`zarr-v3-migration` branch): already on `zarr>=3.1.3`, using v3 APIs
- LINDI: pinned to `zarr>=2.16.1,<3`, still fully v2

**Goal:** Define a shared convention so that LINDI can produce Zarr v3 files that `NWBZarrIO` reads natively, eliminating the h5py shim.

**Key finding:** `object_id` / `source_object_id` are HDMF attributes on groups/datasets — they are never used for reference resolution (only `source` + `path` are used). They are dropped from the unified ref format.

**Cross-reference:** NeurodataWithoutBorders/lindi — a corresponding issue will be filed there.

**Reference implementation:** [zindi](https://github.com/bendichter/zindi) implements this convention for generating Zarr v3 reference file systems from HDF5 files.

---

## Side-by-Side Comparison

| Feature | hdmf-zarr (current) | LINDI (current) |
|---------|---------------------|-----------------|
| **Attribute prefix** | `zarr_` (`zarr_dtype`, `zarr_link`) | `_` prefix (`_SCALAR`, `_COMPOUND_DTYPE`, `_REFERENCE`, `_SOFT_LINK`) |
| **Links** | Parent-centric: `zarr_link` array on parent group with `{"source", "path", "name", "object_id", "source_object_id"}` | Child-centric: empty group with `_SOFT_LINK = {"path": "..."}` attribute |
| **Refs in attrs** | `{"zarr_dtype": "object", "value": {"source", "path", "object_id", "source_object_id"}}` | `{"_REFERENCE": {"source", "path", "object_id", "source_object_id"}}` |
| **Ref resolution** | Only uses `source` + `path` (object_id/source_object_id written but never read) | Same — only `source` + `path` used |
[...]

PR #336, opened 2026-03-20
Adopt unified Zarr v3 convention for hdmf-zarr/Zindi interop

bendichter:
Implements the hdmf-zarr side of the unified Zarr v3 convention for interoperability with Zindi and LINDI (#335, NeurodataWithoutBorders/lindi#125).

- `zarr_link` is replaced by `_LINKS` on the parent group, with entries reduced to `name`, `source`, and `path`.
- `zarr_dtype` is replaced by `_DTYPE`, `_SCALAR`, and `_REFERENCE_FIELDS`. Compound types use the native zarr v3 `structured` data_type.
- References in attributes are stored as `{"_REFERENCE": {"source", "path"}}`. References in datasets are stored as plain target path strings.
- `ZarrReference` drops `object_id` and `source_object_id`, which were written but never used for resolution.
- `NaN`, `Infinity`, and `-Infinity` attribute values are encoded as strings.
- The legacy `zarr_link` and `zarr_dtype` attributes are still read so that Zarr v2 files open with `ZarrV2IO`. The old definitions are documented in a new "Legacy Zarr v2 Convention" section of storage.rst.
- Adds `tests/unit/test_zindi_interop.py`, which generates a Zarr v3 reference file system from an HDF5 NWB file with [zindi](https://github.com/bendichter/zindi) and reads it with `NWBZarrIO`.

#337 no longer reproduces on `zarr-v3-migration` with hdmf 6.2.0, so the `shape` override from #338 is not included.

https://claude.ai/code/session_01XDDwffMD7rJokcon3oyx9Z


docs/source/storage.rst
```diff
@@ -140,12 +141,18 @@ added on any Group or Dataset in the file.
...
+    ``_DTYPE``                    Attribute on Datasets used to specify the data type. Set to
+                                  ``"object_reference"`` for reference datasets. See :ref:`sec-zarr-storage-references`.
+    ``_REFERENCE_FIELDS``         Attribute on compound Datasets listing field names that contain object references.
+    ``_SCALAR``                   Boolean attribute on Datasets (``true``) indicating the dataset holds a scalar value.
     ============================  ======================================================================================
 
+.. note::
+
+    For backward compatibility, the old attribute names ``zarr_link`` and ``zarr_dtype`` are still
+    recognized on read with a deprecation warning.
```
oruebel: Since V3 support is not released yet, I assume V3 should never write with the old attribute conventions. Does this also change how the `backend_V2` writes its data. I assume Yes, since the `ZarrV2IO` inherits from `ZarrIO` (V3). If so, then I am wondering:
- Should we just remove the warnings since the main intent of `ZarrV2IO` is to read legacy data so warning that legacy data has deprecated keys may be redundant?
- Is updating the format for V2 files an issue? I.e., current data is generated with the current keys and then new V2 would be generated with the new keys. It's probably not an issue, but figured its worth mentioning. Or should we make `ZarrV2IO` a read-only?

docs/source/storage.rst
```diff
@@ -140,12 +141,18 @@ added on any Group or Dataset in the file.
...
     ============================  ======================================================================================
-    zarr_link                     Attribute used to store links. See :ref:`sec-zarr-storage-links` for details.
-    zarr_dtype                    Attribute used to specify the data type of a dataset. This is used to implement the
-                                  storage of object references as part of datasets.
-                                  See :ref:`sec-zarr-storage-references`
+    ``_LINKS``                    Attribute on Groups used to store links. See :ref:`sec-zarr-storage-links`.
+    ``_DTYPE``                    Attribute on Datasets used to specify the data type. Set to
+                                  ``"object_reference"`` for reference datasets. See :ref:`sec-zarr-storage-references`.
+    ``_REFERENCE_FIELDS``         Attribute on compound Datasets listing field names that contain object references.
+    ``_SCALAR``                   Boolean attribute on Datasets (``true``) indicating the dataset holds a scalar value.
```
oruebel, line 148: I think it would be nice to preserve the original documentation somehow for reference in case we need to debug old data etc..  We could either add some form of custom note e.g, 

```
.. admonition:: Legacy Data Note
   :class: note
```

here to include the old definitions. Or maybe it would be simpler to create a tagged version of the docs in ReadTheDocs and then have the note simply state that this behavior has changed for Zarr V3 and that for legacy V2 behavior they should look there. What do you think?

oruebel (changes requested):
@bendichter in general I think these updates are fine. Since V3 is already a breaking change, I think this should be included with the V3 release. Since the goal of this PR is Zindi interoperability, is there a test we can add to confirm that the changes here achieve that goal? I also added a few questions related to the docs, but overall I think this makes sense.

docs/source/storage.rst
```diff
@@ -336,33 +323,22 @@ The mappings of data types is as follows
 
 .. note::
 
-    In zarr v3, string and reference fields within compound dtypes are stored as fixed-length
-    Unicode strings (``FixedLengthUTF32``). The string length is dynamically sized to fit the
-    actual data, with a minimum of :py:attr:`~hdmf_zarr.backend.COMPOUND_DTYPE_MIN_STRING_LENGTH`
-    characters to allow for appending rows with longer values. Reference fields are stored as
-    JSON-serialized strings within these fixed-length fields.
+    Compound data types use zarr v3's native ``structured`` data_type, which carries full field
+    information (names and types). No ``_COMPOUND_DTYPE`` attribute is needed.
```
rly:
```suggestion
    Compound data types use zarr v3's native ``structured`` data_type, which carries full field
    information (names and types).
```

rly:
@bendichter could you please update the PR description with the latest? It references `_COMPOUND_DTYPE` which has since been removed.

rly:
I haven't looked closely but this looks good to me.

bendichter:
I merged zarr-v3-migration into this branch and went through the comments.

ZarrV2IO is read-only on zarr-v3-migration, so the old attribute names are only ever read, never written. I removed the deprecation warnings for them, since every v2 file triggers them and there is nothing for the user to do about it. I added a "Legacy Zarr v2 Convention" section at the end of storage.rst with the old definitions and linked to it from the reserved attributes note, which seemed simpler to maintain than a separate tagged docs build.

For the interoperability test, I added test_zindi_interop.py. It writes an NWB file with pynwb, generates a Zarr v3 reference file system from it with zindi, and opens that store with NWBZarrIO, checking scalars, references in attributes and datasets, links, ragged columns, and strings. zindi is not on PyPI (the package with that name there is unrelated), so it is pinned to a commit in the test dependency group and the test skips if it is not installed.

The PR description is updated.

src/hdmf_zarr/backend.py
```diff
@@ -2040,15 +2099,47 @@ def __read_dataset(self, zarr_obj, name):
...
+                "Use zarr v3 structured data_type with _REFERENCE_FIELDS instead." % str(name)
+            )
+
+        compound_dtype = None
+        if hasattr(zarr_obj, "dtype") and hasattr(zarr_obj.dtype, "names") and zarr_obj.dtype.names is not None:
+            # Reconstruct compound dtype descriptor from zarr v3 structured data_type
+            compound_dtype = []
+            for field_name in zarr_obj.dtype.names:
+                if ref_fields and field_name in ref_fields:
+                    compound_dtype.append({"name": field_name, "dtype": "object_reference"})
```
rly, line 2119: Emitting `"object_reference"` here breaks the default `NWBZarrIO.export` for any file containing a compound reference column, which includes `epochs`, `trials`, and the icephys tables. The compound *write* path was not widened to accept the new name, so the descriptor produced by `__read_dataset` cannot be consumed by `write_dataset`.

MWE:

```python
import numpy as np, shutil
from datetime import datetime
from pynwb import NWBFile, TimeSeries

for p in ("a.zarr", "b.zarr"):
    shutil.rmtree(p, ignore_errors=True)

nwb = NWBFile(session_description="d", identifier="i",
              session_start_time=datetime(2024, 1, 1, tzinfo=tzlocal()))
ts = TimeSeries(name="ts", data=np.arange(10.0), unit="v", rate=1.0)
nwb.add_acquisition(ts)
nwb.add_epoch(0.0, 1.0, timeseries=[ts])
nwb.add_epoch(1.0, 2.0, timeseries=[ts])

with NWBZarrIO("a.zarr", mode="w") as io:
    io.write(nwb)

with NWBZarrIO("a.zarr", mode="r") as r:
    with NWBZarrIO("b.zarr", mode="w") as w:
        w.export(src_io=r)
```

On `zarr-v3-migration` (af58284d) this exports cleanly. On this branch:

```
  File "src/hdmf_zarr/backend.py", line 1492, in write_dataset
    arr = np.array(new_items, dtype=dtype)
TypeError: float() argument must be a string or a real number, not 'ZarrReference'
```

It reproduces only with the default `link_data=True`. Passing `write_args={"link_data": False}` sets `clear_cache=True` (backend.py:592-593), which sends `HDMFIO.export` through the BuildManager, where the reference field's dtype arrives from the spec as a dict (`{'target_type': 'TimeSeries', 'reftype': 'object'}`) and is caught by the existing `isinstance(field["dtype"], dict)` branches. The failing route is `bldr = src_io.read_builder()` (hdmf/backends/io.py:177), where the dtype is the one written here. That is also why the #337 reproducer, which passes `link_data=False`, does not hit this.

Cause: two consumers of `options["dtype"]` in the same block disagree about the vocabulary. Detection at line 1451 goes through `_is_ref`, which this PR widened to accept `"object_reference"` (line 976), so `refs` is populated and `_create_ref` puts real `ZarrReference` objects into the rows. The numpy dtype construction at lines 1476-1489 falls through to `__resolve_dtype_helper__`, and `__dtypes` (line 1640) has `"object"` but no `"object_reference"`, so it resolves to `None`. `np.dtype([("idx_start", "<i4"), ("count", "<i4"), ("timeseries", None)])` silently becomes `float64`, and the cast at line 1492 then tries to coerce a `ZarrReference` to a float. Fresh writes are unaffected because there the descriptor comes from the spec and still says `"object"` or the dict form. Only the read path emits the bare string.

Two lines fix it:

```diff
@@ __dtypes, backend.py:1640
         "object": ZarrReference,
+        "object_reference": ZarrReference,
     }
```

```diff
@@ write_dataset, backend.py:1501
                     is_ref_field = (
-                        field["dtype"] == "object" or isinstance(field["dtype"], dict)
+                        field["dtype"] in ("object", "object_reference") or isinstance(field["dtype"], dict)
                     )
```

With both applied the MWE exports cleanly and the suite is still green (551 passed, 1 skipped, ignoring the network dependent `test_fsspec_streaming.py`).

Worth a regression test, and it needs to be a new one. The compound reference export cases at `tests/unit/test_io_convert.py:743-877` cannot just be uncommented: they are missing the `BazCpdData` import, and once that is added they still fail identically on `zarr-v3-migration` for a separate reason (the built `BazCpdData` builder comes back with `dtype=int` rather than the compound list from its spec, so it never reaches the compound branch at all). The MWE above is a better basis.

src/hdmf_zarr/zarr_utils.py
```diff
@@ -198,6 +203,13 @@ def types(self):
     def dtype(self):
         return self.__dtype
 
+    @property
+    def shape(self):
+        # Don't expose raw zarr shape — compound table datasets should never
+        # be treated as scalars by hdmf's objectmapper (which unwraps data
+        # when shape == (1,) and data[0] is not np.void).
+        return None
+
```
rly, line 212: #337 no longer reproduces on `zarr-v3-migration` with hdmf 6.2.0, which is already the floor in `pyproject.toml`.

The override also breaks `pynwb.validate` for any file with an epochs, trials, or icephys table, since hdmf does `len(np.shape(builder.data))` at `validate/validator.py:553` and `np.shape` reads this property directly:

```
TypeError: object of type 'NoneType' has no len()
```

Removing it, those files validate with 0 errors and the suite is unchanged (552 passed either way).

```suggestion
```

alejoe91 (commented):
LGTM after changes proposed by @rly

src/hdmf_zarr/backend.py
```diff
@@ -2040,15 +2099,47 @@ def __read_dataset(self, zarr_obj, name):
...
+        # Resolve dtype from new unified convention attributes, with backward compat for old names
+        is_scalar = zarr_obj.attrs.get("_SCALAR", False)
+        dtype_attr = zarr_obj.attrs.get("_DTYPE", None)
+        ref_fields = zarr_obj.attrs.get("_REFERENCE_FIELDS", None)
+
+        if zarr_obj.attrs.get("_COMPOUND_DTYPE", None) is not None:
+            raise ValueError(
+                "_COMPOUND_DTYPE attribute is no longer supported on dataset '%s'. "
+                "Use zarr v3 structured data_type with _REFERENCE_FIELDS instead." % str(name)
+            )
```
rly, line 2111:
```suggestion
```

`_COMPOUND_DTYPE` only existed in intermediate commits in this PR

src/hdmf_zarr/backend.py
```diff
@@ -971,33 +1016,43 @@ def resolve_ref(self, zarr_ref):
...
+                # get rid of extra "/" and "./" in the path root and source_file
+                root_path = str(self.path).rstrip("/")
+                source_path = str(source_file).lstrip(".")
+                source_file = root_path + source_path
+
+            target_zarr_obj = self._open_file_consolidated(
+                store=source_file,
+                mode="r",
+                storage_options=self.__storage_options,
+            )
```
rly, line 1055: This branch looks unreachable. The guard at line 1007 already returns for `source == "."`, so `is_same_file` can only be true when the reference dict has no `source` key at all, and nothing produces one: `_create_ref` always sets it, the plain path fallback at line 1001 synthesizes `{"source": "."}`, and zindi writes `"source": "."` explicitly. I instrumented the body and ran the suite including `test_zindi_interop.py`; it is never entered.

Same-file references over a non-`LocalStore` store are already handled by the 1007 guard, which also reuses `self.__file`.

```suggestion
        if source is None:
            source_file = str(zarr_ref["path"])
        else:
            source_file = str(source)

        if not self.is_remote():
            if isinstance(self.source, str) and self.source.startswith(("s3://")):
                source_file = self.source
            else:
                source_file = self._resolve_ref_source(source_file)
        else:
            # get rid of extra "/" and "./" in the path root and source_file
            root_path = str(self.path).rstrip("/")
            source_path = str(source_file).lstrip(".")
            source_file = root_path + source_path

        target_zarr_obj = self._open_file_consolidated(
            store=source_file,
            mode="r",
            storage_options=self.__storage_options,
        )
```

src/hdmf_zarr/backend.py
bendichter: Thanks for tracking this down. I applied both changes and added a test in test_nwbzarrio.py that exports a file with two epochs using the default link_data=True and checks the exported compound dataset. The test checks that dataset with zarr directly rather than reading the whole export back through NWBZarrIO, because that read fails on the root-level file_create_date dataset, which the export writes as a self-link to a target that does not exist in the new store. That also happens on zarr-v3-migration, so I left it out of this PR.

src/hdmf_zarr/zarr_utils.py
bendichter: Removed. I confirmed that a one-row epochs table written directly and one exported from HDF5 with link_data=False both read correctly without it, and added a test that pynwb.validate returns no errors for a file with a compound reference column. I also updated the PR description so it no longer claims to fix #337.

src/hdmf_zarr/backend.py
bendichter: Removed, along with the comment in write_dataset that mentioned it.

src/hdmf_zarr/backend.py
bendichter: Applied. The suite passes without it.

oruebel (approved):

PR #338, opened 2026-03-21
Fix single-row compound datasets with references being flattened

### LINKED ISSUE #337: BuilderZarrTableDataset flattens single-row compound datasets with references
## Bug

When reading a compound dataset with reference fields that contains exactly 1 row, `BuilderZarrTableDataset` returns a flat row `[val1, val2, <GroupBuilder>]` instead of a list of rows `[[val1, val2, <GroupBuilder>]]`. This causes `DynamicTable.__init__` to fail with:

```
ValueError: Must provide same number of ids as length of columns
```

because `len([val1, val2, <GroupBuilder>])` = 3 (number of fields) instead of 1 (number of rows).

## Reproduction

```python
from pynwb import NWBFile, NWBHDF5IO, TimeSeries
from pynwb.icephys import VoltageClampStimulusSeries, VoltageClampSeries, IntracellularElectrode
from datetime import datetime

# Run the icephys tutorial (plot_icephys.py) to generate ex_test_icephys_file.nwb
# Then:
with NWBHDF5IO("ex_test_icephys_file.nwb", "r") as read_io:
    with NWBZarrIO("ex_test_icephys_file.nwb.zarr", "w") as export_io:
        export_io.export(src_io=read_io, write_args={"link_data": False})

with NWBZarrIO("ex_test_icephys_file.nwb.zarr", "r") as io:
    nwb = io.read()  # raises ConstructError
```

## Root cause

In `zarr_utils.py`, `AbstractZarrTableDataset.__getitem__` with a slice on a 1-element structured array. The `DatasetOfReferences` base class `__getitem__` may return data that, when iterated, produces a single row that gets stored as `data` in the `DatasetBuilder` without the outer list wrapper.

## Affected tables

Any NWB table with compound reference columns and exactly 1 row, including:
- `IntracellularStimuliTable.stimuli` (TimeSeriesReferenceVectorData)
- `IntracellularResponsesTable.responses` (TimeSeriesReferenceVectorData)

## Context

Found while testing the unified Zarr v3 convention (#335, #336) with the PyNWB icephys tutorial. The bug exists on the `zarr-v3-migration` branch and is not specific to the convention changes.

bendichter:
## Summary
Override `shape` property on `AbstractZarrTableDataset` to return `None` instead of the raw zarr array shape. This prevents hdmf's `objectmapper` from treating single-row compound datasets as scalars and unwrapping `data[0]` into a flat row.

## Root cause
hdmf's `ObjectMapper.__get_subspec_values` unwraps `builder.data` when `shape == (1,)` and `data[0]` is not `np.void`. For `BuilderZarrTableDataset`, this replaces the lazy wrapper with a flat row `[val1, val2, GroupBuilder]`, causing a column length mismatch in `DynamicTable.__init__`.

## Fix
Return `None` for `shape` on compound table datasets so the scalar unwrap is never triggered.

## Test plan
- [x] Extended icephys tutorial reads via both Zarr conversion and zindi
- [x] All 448 tests pass

Fixes #337

### DIFF
--- a/src/hdmf_zarr/backend.py
+++ b/src/hdmf_zarr/backend.py
@@ -1559,7 +1559,6 @@ class ZarrIO(HDMFIO):
 
     __reserve_attribute = (
         "_DTYPE", "_SCALAR", "_LINKS", "_REFERENCE_FIELDS",
-        "_COMPOUND_DTYPE",  # backward compat (no longer written)
         "zarr_dtype", "zarr_link",  # backward compat with old convention
         SPEC_LOC_ATTR,
     )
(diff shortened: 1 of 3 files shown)


oruebel:
@rly could you take a look at this. Is this something that should be fixed in HDMF or is this approach appropriate?

oruebel:
@bendichter is this still needed?

bendichter:
It depends on the minimum hdmf version we want to support. I retested this with the current zarr-v3-migration code: on hdmf 4.3.1 (pynwb 3.1.3) the single-row read from #337 still fails with the ConstructError, and this PR fixes it. On hdmf 6.2.0 (pynwb 4.2.0) the read works without the PR.

rly:
I think it's fine for hdmf-zarr to support only the latest hdmf, 6.2.0. The zarr v3 branches require the latest hdmf right now.

PR #348, opened 2026-05-14
Fix `resolve_ref` self-reference resolution over fsspec stores

h-mayorquin:
`ZarrIO.resolve_ref` raises `PathNotFoundError: nothing found at path ''` when resolving any internal reference whose `source` is `"."` over an fsspec-backed store. `"."` is the convention `ZarrIO` itself emits for every same-file reference (Device, ElectrodeGroup, every `DynamicTableRegion` target, every cached spec link), so this error fires on essentially any non-trivial NWB Zarr file read over HTTPS, S3, GCS, or any other fsspec scheme. The bug is that `resolve_ref` re-opens the URL of the current store as if it were external: that works for local paths because `os.path.join(self.source, ".")` resolves correctly, but on fsspec stores `"."` collapses to an empty key lookup. The fix is a roughly ten-line short-circuit at the top of `resolve_ref`: when `source == "."`, reuse the already-open `self.__file` and skip `__open_file_consolidated` entirely. The existing external-source path below is untouched, so cross-file references keep working exactly as before.

Verified end-to-end against `https://dandiarchive.s3.amazonaws.com/zarr/c8c6b848-fbc6-4f58-85ff-e3f2618ee983/` (the DANDI 000719 file used by this repo's PR #330 S3 streaming tutorial); the read now returns the expected `NWBFile` and a new regression test in `test_fsspec_streaming.py` covers it. 

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -150,3 +150,6 @@ _version.py
 .ruff_cache
 zarr_test*.zarr
 *.nwb.zarr
+
+# uv
+uv.lock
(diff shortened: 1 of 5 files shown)


rly:
Makes sense. Thanks for the PR @h-mayorquin ! I added a few tests to increase code/branch coverage.

rly (approved):

PR #349, opened 2026-05-29
Read zarr v2 NWB files via a dedicated `NWBZarrV2IO`

alejoe91:
That PR baked the zarr v2 fallbacks directly into `ZarrIO` / `NWBZarrIO` — so every v3 read went through `try / except / use_consolidated=False`, every group walk used a chunk-decoding iterator, and ~440 lines of v2-specific code lived in `backend.py`. It worked but the v3 path paid the complexity tax for files it would never see.

This PR splits the two:

- `ZarrIO` and `NWBZarrIO` go back to their clean v3 shape.
- A new `ZarrV2IO(ZarrIO)` (read-only) and `NWBZarrV2IO(ZarrV2IO)` live in `backend_zarrv2.py` / `nwb_zarrv2.py` and contain all the v2 logic — consolidated-metadata fallbacks, raw-chunk decoding for `|O` arrays, dual path resolution for old-style references, the spec-reader fallback.
- To make this possible, `ZarrIO` grew a few protected hooks (`_open_file`, `_open_file_consolidated`, `_iter_children`, `_resolve_ref_source`, `_make_spec_reader`, `_open_for_namespaces`). No behavior change for v3 users; `ZarrV2IO` overrides them.

### Usage

```python
from hdmf_zarr import NWBZarrIO, NWBZarrV2IO

# explicit:
with NWBZarrV2IO("old_v2_file.nwb.zarr", mode="r") as io:
    nwbfile = io.read()

# read_nwb zarrv2
nwbfile = NWBZarrV2IO.read_nwb("old_v2_file.nwb.zarr")
```

`NWBZarrV2IO` is read-only by design — writes always produce v3.

The `NWBZarrV2IO` has helper functions to export/convert to Zarr v3:
```python
io = NWBZarrV2IO("old_v2_file.nwb.zarr", mode="r")

io.export_to_v3(output="new_v3_file.nwb.zarr")

# or directly
NWBZarrV2IO.convert_to_v3(source_path="old_v2_file.nwb.zarr", dest_path="new_v3_file.nwb.zarr")
```

### Examples

Updated `plot_s3_streaming.py` to use the NWBZarrV2IO reader.

### Tests

- `test_zarrv2_backward_compat.py` now uses `NWBZarrV2IO`.
- The remote S3 v2 fixtures in `test_fsspec_streaming.py` are switched to `NWBZarrV2IO`
- Three `test_zarrio.py` tests that poked `_ZarrIO__*` mangled attributes were updated to the new protected names.


alejoe91:
@bendichter @oruebel let me know what you guys think about this approach

src/hdmf_zarr/v2_backend.py
```diff
@@ -0,0 +1,476 @@
+"""Read-only backend for NWB / HDMF Zarr files written with hdmf-zarr<1.0 + zarr<3.
```
oruebel:
```suggestion
"""Read-only backend for NWB / HDMF ZarrV2 files written with hdmf-zarr<1.0 + zarr<3.
```

src/hdmf_zarr/backend_zarrv2.py
```diff
@@ -0,0 +1,476 @@
+"""Read-only backend for NWB / HDMF Zarr files written with hdmf-zarr<1.0 + zarr<3.
+
+Zarr-python v3 cannot fully parse some constructs produced by zarr v2 (object
```
oruebel, line 3: For consistency, I would rename this file to `backend_v2.py`, i.e., have the version as a post-fix rather than a prefix

oruebel:
> That PR baked the zarr v2 fallbacks directly into `ZarrIO` / `NWBZarrIO` — so every v3 read went through `try / except / use_consolidated=False`, every group walk used a chunk-decoding iterator, and ~440 lines of v2-specific code lived in `backend.py`. It worked but the v3 path paid the complexity tax for files it would never see.

I think the general approach of separating support for V2 files from the V3 backend makes sense. The overall code may be a bit longer right now but it'll make maintenance easier and keep the V3 code cleaner and more focused. I'll take a closer look at the actual code in the next couple of days.

.github/workflows/test_backward_compat.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: Generate zarr v2 NWB test file
+        run: |
+          /tmp/zarr_v2_env/bin/python tests/unit/helpers/generate_v2_nwb_zarr.py \
+            --nwb-output-path tests/unit/helpers/v2_test_file.nwb.zarr \
+            --expectations-output-path tests/unit/helpers/v2_expected.json
+          echo "--- generated file tree ---"
+          find tests/unit/helpers/v2_test_file.nwb.zarr -maxdepth 2 -type d | head -30
+          echo "--- expected metadata ---"
+          cat tests/unit/helpers/v2_expected.json
```
oruebel: Since the file for testing should be small, I am wondering whether it would be easier to add the V2 test files to either this or a separate repo. Generating the files this way works right now, but as Python and library versions changes I suspect that this may become brittle. If the files are more readily available, then the V2 tests could potentially even run as part of the regular tests, although, I think in the long-run it makes sense to have this as a separate test suite.

src/hdmf_zarr/backend_zarrv2.py
```diff
@@ -0,0 +1,476 @@
...
+supported — files written by this codebase are always zarr v3.
+"""
+
+import json
+import os
+import warnings
+
+import numpy as np
+import zarr
+from zarr.storage import LocalStore
```
oruebel, line 16: @alejoe91 just to clarify, the V2 backend uses Zarr V3 to read V2 files, i.e., there are no additional dependencies needed to support V2 read?

.github/workflows/test_backward_compat.yml
oruebel: @rly what do you think?

src/hdmf_zarr/__init__.py
```diff
@@ -12,7 +14,7 @@
     __version__ = version("hdmf")
     del version
 
-__all__ = ["ZarrIO", "ZarrDataIO", "NWBZarrIO"]
+__all__ = ["ZarrIO", "ZarrDataIO", "NWBZarrIO", "ZarrV2IO", "NWBZarrV2IO", "is_zarr_v2_file"]
```
oruebel, line 17: If supporting ZarrV2 does not require significant additional dependencies, then I think always exposing both makes sense. Otherwise, we may want to consider whether V2 should be an optional dependency?

src/hdmf_zarr/backend.py
```diff
@@ -192,10 +192,10 @@ def __init__(self, **kwargs):
         self._written_builders = WriteStatusTracker()  # track which builders were written (or read) by this IO object
         self.__dci_queue = None  # Will be initialized on call to io.write
         # Cache for consolidated metadata to avoid repeated opening of the same files
-        self.__consolidated_cache = {}
+        self._consolidated_cache = {}
```
oruebel, line 222: @alejoe91 it looks like one main change here is to make a number of functions and variables protected rather than private (i.e., replace double-underscore with a single underscore). What is the reason for that change?

oruebel: I just noticed that `ZarrV2IO` inherits from `ZarrIO` so changing methods to protected instead makes sense.

src/hdmf_zarr/backend.py
```diff
@@ -1719,11 +1735,32 @@ def __get_built(self, zarr_obj):
...
         return self.__built.get(path, None)
 
+    def _resolve_ref_source(self, source_file):
+        """Resolve the *source_file* of a Zarr reference to an absolute filesystem path.
+
+        Hook for subclasses to alter the resolution policy. The default resolves
+        *source_file* relative to ``self.source``: this matches what the current
+        hdmf-zarr writer produces (self-references stored as ``"."``).
+        """
+        return os.path.abspath(os.path.normpath(os.path.join(self.source, source_file)))
```
oruebel, line 1918: Just checking, does this path logic also work for files on S3?

src/hdmf_zarr/backend_zarrv2.py
```diff
@@ -0,0 +1,476 @@
...
+        resolved_from_parent = os.path.abspath(os.path.normpath(os.path.join(base_dir, source_file)))
+        if os.path.isdir(resolved_from_parent):
+            return resolved_from_parent
+        return abs_source
+
+    # ----- group iteration with v2 chunk fallback -----
+
+    def _iter_children(self, zarr_obj):
+        store = zarr_obj.store
+        group_prefix = zarr_obj.path or ""
```
oruebel, line 499: Could you add a docstring here and as approbriate inline comments to describe the logic of this function. In `backend.py` you mentioned that this function may return a `DatasetBuilder` instead in the case of the V2 backend, could you clarify here why this is necessary.

src/hdmf_zarr/backend.py
```diff
@@ -1736,24 +1773,25 @@ def __read_group(self, zarr_obj, name=None, ignore_groups=set()):
...
-            sub_builder = self.__read_group(sub_group, sub_name)
-            ret.set_group(sub_builder)
-
-        # read sub datasets
-        for sub_name, sub_array in zarr_obj.arrays():
-            sub_builder = self.__read_dataset(sub_array, sub_name)
-            ret.set_dataset(sub_builder)
+        for sub_name, child in self._iter_children(zarr_obj):
+            if isinstance(child, DatasetBuilder):
+                # Subclasses may yield pre-built dataset builders (e.g. v2 fallback)
```
oruebel: Why is `DatasetBuilder` needed as an alternative? And separately,  why is `GroupBuilder` not also needed here? I think it would be useful to clarify this in a comment in the code

src/hdmf_zarr/nwb.py
```diff
@@ -81,20 +92,21 @@ def export(self, **kwargs):
     )
     def read_nwb(**kwargs):
         """
-        Helper factory method for reading an NWB file and return the NWBFile object
+        Helper factory method for reading an NWB file and return the NWBFile object.
+
+        Sniffs the layout and automatically dispatches to :class:`NWBZarrV2IO` when
+        *path* points to a zarr v2 hierarchy.
```
oruebel: To be honest, I found it confusing at first that a function `NWBZarrIO` would return a `NWBZarrV2IO` instance. Also having a static `read_nwb` method in both `NWBZarIO` and `NWBZarrV2IO` seems redundant. With that, I am wondering whether it may be cleaner to have a separate, stand-alone function `read_nwb_zarr` for this purpose to delegate between V2 and V3, rather than adding V2 logic in the V3 class. Optionally, we could then remove the separate `NWBZarrIO.read_nwb` and `NWBZarrV2IO.read_nwb` functions. @rly thoughts?

oruebel: We should also check if `read_nwb` in PyNWB https://github.com/NeurodataWithoutBorders/pynwb/blob/bd9bcc275a17af73dd509b813764341706539dc5/src/pynwb/__init__.py#L573 will need to be updated to support the change here. I think with the current approach (where NWBZarrIO falls back to NWBZarrV2IO `NWBZarrIO.read_nwb` ) the `read_nwb` function may still work correctly.

src/hdmf_zarr/nwb_zarrv2.py
```diff
@@ -0,0 +1,66 @@
+"""Read-only NWB IO backend for NWB Zarr files written with hdmf-zarr<1.0 + zarr<3.
+
```
oruebel: I think it would be Ok for this code to also live in `nwb.py`, in particular since `nwb.py` imports this code anyways. However, either way is fine with me.

src/hdmf_zarr/v2_backend.py
```diff
@@ -0,0 +1,476 @@
...
+    async def _list():
+        result = []
+        async for item in store.list_dir(prefix):
+            result.append(item)
+        return result
+
+    return sorted(zarr_sync(_list()))
+
+
+def _decode_v2_chunk(raw, compressor, filters, dtype, chunk_shape, order, is_object):
```
oruebel: This function, as well as a number of other functions in this file, have several parameters and it is not necessarily clear what they all are. It would be helpful to add docstrings for the parameters and ideally also add type hints.

src/hdmf_zarr/v2_backend.py
```diff
@@ -0,0 +1,476 @@
...
+        return os.path.exists(os.path.join(str(store.root), key))
+    from zarr.core.sync import sync as zarr_sync
+
+    async def _exists():
+        return await store.exists(key)
+
+    return zarr_sync(_exists())
+
+
+def _store_list_dir(store, prefix):
```
oruebel: It looks like most of these functions may only be used internally by `ZarrV2IO`. If so, I think it may be cleaner to make this private, static methods of `ZarrV2IO` instead.

oruebel:
Overall the approach looks good. I still need to take a closer look at `v2_backend.py` and the tests, but added comments to the other files.

src/hdmf_zarr/backend_zarrv2.py
alejoe91: yes that's correct

src/hdmf_zarr/__init__.py
alejoe91: no additional dependency needed!

src/hdmf_zarr/backend.py
alejoe91: thanks, I'll double check

alejoe91:
Thanks for the review @oruebel 

I'll tackle them in the next days but happy to hear that the approach is fine!

src/hdmf_zarr/__init__.py
oruebel: Great

src/hdmf_zarr/v2_backend.py
alejoe91: moved all these helper functions as static methods in last commit

src/hdmf_zarr/v2_backend.py
alejoe91: added docstrings in last commit

.github/workflows/test_backward_compat.yml
alejoe91: that could be a good option. I can commit the generated files (currently 1.2M, but it can be made smalleR). Let me know what you decide.

src/hdmf_zarr/backend.py
alejoe91: this function is oly reached for non-remote paths: https://github.com/alejoe91/hdmf-zarr/blob/nwb-zarr-v2-io/src/hdmf_zarr/backend.py#L936-L940

src/hdmf_zarr/backend.py
alejoe91: Added comment in last commit

src/hdmf_zarr/nwb.py
alejoe91: yeah that's a good point. Should I make a standalone `read_nwb_zarr`? For `pynwb` integration, maybe both `hdmf` and `hdmf_zarr` should have such a `read_nwb` function and `pynwb` should just pick which package to use based on the backend (`hdmf.read_nwb` VS `hdmf_zarr.read_nwb`)

src/hdmf_zarr/nwb_zarrv2.py
alejoe91: happy to move it if you feel it's better. I kept it separate because it makes it clear that it's a separate reader/backend (and might be dropped in the future)

alejoe91:
@oruebel @rly I tackled all the comments in the last commit. Last things we have to decide:

- how to deal with `read_nwb` functions: https://github.com/hdmf-dev/hdmf-zarr/pull/349#discussion_r3454705296
- whether to merge everything into same files (I'd prefer not to, but I'm open): https://github.com/hdmf-dev/hdmf-zarr/pull/349#discussion_r3454743843

src/hdmf_zarr/nwb_zarrv2.py
oruebel: I don't have a strong preference either way. I think its fine to keep as is.

src/hdmf_zarr/nwb_zarrv2.py
```diff
@@ -0,0 +1,169 @@
+"""Read-only NWB IO backend for NWB Zarr files written with hdmf-zarr<1.0 + zarr<3.
```
oruebel: I think the `_v2` postfix could be misleading to users, because it is not clear what the version number refers to. As such a user may think that `backend_v2.py` is actually newer than `backend.py` or that `nwb_v2.py maybe be specific to NWBv2 files. I think using the slightly more explicit postfix e.g., `_zarr2` throughout would help.

.github/workflows/test_backward_compat.yml
oruebel: I'm leaning towards adding the test file, bot for simplicity and to simplify maintenance. @rly what do you prefer?

src/hdmf_zarr/backend_zarrv2.py
```diff
@@ -0,0 +1,563 @@
...
+        still understood by zarr v3).
+
+        Parameters
+        ----------
+        store :
+            The zarr store backing the file (local or remote).
+        group_path : str
+            Path within the store of the parent group (empty string for root).
+        name : str
+            Name of the dataset within *group_path*.
```
oruebel: `hdmf_zarr` uses Sphinx-style docstrings. Could you please update accordingly to make sure the docs will render consistently

src/hdmf_zarr/nwb.py
oruebel: > Should I make a standalone `read_nwb_zarr`

I think that makes sense. It will require some small changes to in PyNWB's `read_nwb`, but I think that's fine for the sake of clarity https://github.com/NeurodataWithoutBorders/pynwb/blob/2e38a0c8aa4dfaa682133b7df4d33011670cd8b3/src/pynwb/__init__.py#L573


> `pynwb` should just pick which package to use based on the backend (`hdmf.read_nwb` VS `hdmf_zarr.read_nwb`)

Yes, that is what `read_nwb` in PyNWB currently does, only with the slight difference that it: 1) uses the `read_nwb` methods defined on the I/O classes, i.e., `NWBZarrIO.read_nwb` and `NWBHDF5IO.read_nwb` and 2) `NWBHDF5IO` is implemented in PyNWB and not HDMF. 

> For `pynwb` integration, maybe both `hdmf` and `hdmf_zarr` should have such a `read_nwb` function

NWB-specific logic is usually not included in HDMF, but delegated to PyNWB. As such, `NWBIO` and `read_nwb` are currently defined in PyNWB. We mainly included the NWB-specific code in hdmf_zarr because hdmf_zarr was still under development and was (or still is) only an optional dependency of PyNWB.

src/hdmf_zarr/backend_zarrv2.py
alejoe91: done

src/hdmf_zarr/nwb_zarrv2.py
alejoe91: I renamed the _v2 files to _zarrv2

alejoe91:
@oruebel this is ready on my end. Let me know if you want to address the `read_*` here or once it's merged to #325

src/hdmf_zarr/backend_zarrv2.py
oruebel: @alejoe91 did you already address this?

src/hdmf_zarr/nwb.py
oruebel: Since this will require some changes in PyNWB's `read_nwb` anyways, I think the simplest may be to just delegate the job of deciding which IO backend to use to PyNWB's `read_nwb` method. That function already has to make the decision of whether to use `NWBHDF5IO` or `NWBZarrIO`, and I think it's perfectly fine for it to also make the choice whether to then use `NWBZarrV2IO`. 
I.e., I would propose removing the following code here:

```
        from .backend_zarrv2 import is_zarr_v2_file
        from .nwb_zarrv2 import NWBZarrV2IO

        io_cls = NWBZarrV2IO if is_zarr_v2_file(path, storage_options=storage_options) else NWBZarrIO
        io = io_cls(path=path, mode="r", load_namespaces=True, storage_options=storage_options)
        return io.read()
```

and instead adding the equivalent logic in the PyNWB's `read_nwb` here https://github.com/NeurodataWithoutBorders/pynwb/blob/2e38a0c8aa4dfaa682133b7df4d33011670cd8b3/src/pynwb/__init__.py#L573-L630

docs/source/overview.rst
```diff
@@ -30,7 +30,7 @@ Known Limitations
 -----------------
 
 - The Zarr backend is currently experimental and may still change.
-- **Zarr v3 only:** This version of hdmf-zarr writes zarr v3 format only and cannot read files written in zarr v2 format. To read zarr v2 files, use hdmf-zarr 0.x with zarr-python v2. The `zarr CLI migration tool <https://zarr.readthedocs.io/en/stable/user-guide/cli/>`_ may also be used to convert v2 files to v3 format.
+- **Zarr v3 only for write:** This version of hdmf-zarr writes zarr v3 format only. Zarr v2 folders can only be read in read-only mode with the :py:class:`~hdmf_zarr.nwb_v2.NWBZarrV2IO` class. Zarr v2 folders can be converted to v3 using the :py:meth:`~hdmf_zarr.nwb_v2.NWBZarrV2IO.convert_to_v3` function.
```
oruebel:
```suggestion
- **Zarr v2 read-only:** Zarr v2 folders can only be read in read-only mode with the :py:class:`~hdmf_zarr.backend_zarrv2.ZarrV2IO` and corresponding :py:class:`~hdmf_zarr.nwb_v2.NWBZarrV2IO` classes. Writing of new Zarr v2 files is not supported, i.e., new files can only be written in Zarr v3 via :py:class:`~hdmf_zarr.backend.ZarrIO` and :py:class:`~hdmf_zarr.nwbNWBZarrIO`. Zarr v2 folders can be converted to v3 using the :py:meth:`~hdmf_zarr.nwb_v2.NWBZarrV2IO.convert_to_v3` function. (since v0.14)
```

The `Zarr v3 only for write` sounds like hdmf_zarr may only be able to write v3 but not read. Since the change is to full support or Zarr v3 and limited (read-only_ support for Zarr v2, I would propose to rephrase this to instead state the limitations for v2.

src/hdmf_zarr/backend_zarrv2.py
```diff
@@ -0,0 +1,588 @@
...
+    def _read_v2_dataset(self, store, group_path, name):
+        """Read a zarr v2 dataset by decoding raw chunks, returning a DatasetBuilder.
+
+        Zarr-python v3 raises when trying to open certain zarr v2 arrays — most
+        commonly those with an object dtype stored via ``pickle``, ``json2``, or
+        ``vlen-utf8`` filters, or with a fill-value incompatible with the encoded
+        dtype.  When ``zarr_obj[entry]`` fails in :meth:`_iter_children` for such
+        an entry, this method is called as a fallback: it reads the raw ``.zarray``
+        and ``.zattrs`` metadata directly from the store and decodes every chunk
+        manually via :meth:`_decode_v2_dataset`.
```
oruebel, line 558: It looks like in cases where ZarrV3 can't read the V2 array, this will actually load the full data into memory, instead of reading the data lazily. To clarify this I would propose to:
- Add a comment in the docstring of `ZarrV2IO` to clarify
- Create a follow-up issue to support lazy ready for these datasets if performance should be an issue. I.e., this logic could be pushed into a wrapper class to decode data lazily on request rather than on file read. I think this should be addressed in a separate PR if it becomes necessary, but I think it would be useful to document this in an issue.

tests/unit/test_fsspec_streaming.py
```diff
@@ -1,12 +1,14 @@
...
 from .helpers.utils import check_s3fs_ffspec_installed
 
 HAVE_FSSPEC = check_s3fs_ffspec_installed()
 
 
-@unittest.skip("S3 test files are in zarr v2 format with object_codec which is not supported by zarr v3")
 class TestFSSpecStreaming(unittest.TestCase):
+    """Stream remote NWB Zarr files. The current S3 test fixtures are zarr v2, so
+    these tests instantiate :class:`NWBZarrV2IO` directly; the convenience
+    :meth:`NWBZarrIO.read_nwb` auto-dispatcher is exercised separately below."""
```
oruebel: Does this mean the streaming tests are all testing with Zarr V2 such that streaming read with Zarr V3 is currently not being tested? If so, could you create a follow-up issue for this or maybe this is being addressed already in the V3-migration PR?

tests/unit/helpers/generate_nwb_zarrv2.py
```diff
@@ -0,0 +1,137 @@
+"""Generate a zarr v2 NWB file using hdmf-zarr<1.0 and zarr<3.
```
oruebel: For consistency, I would rename the file to `generate_nwb_zarrv2.py`

tests/unit/test_zarrv2_backward_compat.py
```diff
@@ -0,0 +1,319 @@
+"""Backward compatibility test: read a zarr v2 NWB file with the current (zarr v3) code.
+
```
oruebel, line 2: For consistency, I would rename the file to `test_zarrv2_backward_combat.py`

tests/unit/helpers/generate_nwb_zarrv2.py
oruebel: Just flagging, if this file is being renamed then the docstring in `tests/unit/test_v2_backward_compat.py` will also need to be updated accordingly.

tests/unit/test_v2_backward_compat.py
```diff
@@ -0,0 +1,319 @@
+"""Backward compatibility test: read a zarr v2 NWB file with the current (zarr v3) code.
+
+This test reads a NWB zarr file that was generated with hdmf-zarr<1.0 + zarr<3
+(see ``helpers/generate_v2_nwb_zarr.py``) and validates that key metadata and
```
oruebel:
```suggestion
This test reads a NWB zarr file that was generated with hdmf-zarr<0.14+ zarr v2.x
(see ``helpers/generate_v2_nwb_zarr.py``) and validates that key metadata and
```

@rly do you know what the intended version for the Zarr V3 would be? Just wanted to flag this, since this mentions hdmf-zarr 1.0 and I wasn't sure whether the plan was to raise to a 1.0?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 ### Added
 - Added `ZarrIO.generate_dataset_html` method to generate rich HTML representations of Zarr arrays for display in Jupyter notebooks. @rly [#306](https://github.com/hdmf-dev/hdmf-zarr/pull/306)
 - Added S3 streaming tutorial with consolidated metadata guidance. @rly @oruebel [#308](https://github.com/hdmf-dev/hdmf-zarr/pull/308) [#330](https://github.com/hdmf-dev/hdmf-zarr/pull/330)
+- Added `NWBZarrV2IO.export_to_v3` and the one-shot `NWBZarrV2IO.convert_to_v3` static helper to convert NWB Zarr v2 files to Zarr v3. @alejoe91 [#XXX](https://github.com/hdmf-dev/hdmf-zarr/pull/XXX)
```
oruebel: @alejoe91 this PR seems to add number of other major changes. Are those already covered by the Changelog of the parent PR? Just wanted to double-check.

oruebel:
> @oruebel this is ready on my end. Let me know if you want to address the `read_*` here or once it's merged to #325

Please see the corresponding thread above regarding  `read_*`. Since this is being merged into another branch (rather than dev), I think it's fine to address in either place, but I think it's a simple enough change that it can probably be done here, but up to you. 

@alejoe91 I added a few additional comments. Nothing major, just a few minor items that should hopefully pretty simple to address. 

From my end this looks fine. @rly @bendichter I'll leave final approvals for this to you.

src/hdmf_zarr/nwb.py
alejoe91: ok, let me remove the logic from here

src/hdmf_zarr/backend_zarrv2.py
alejoe91: Added a comment in docstring and opened issue here: https://github.com/hdmf-dev/hdmf-zarr/issues/356

tests/unit/test_zarrv2_backward_compat.py
alejoe91: done

tests/unit/helpers/generate_nwb_zarrv2.py
alejoe91: done

src/hdmf_zarr/nwb.py
alejoe91: done

CHANGELOG.md
alejoe91: I rephrased it: the main contribution is adding the `NWBZarrV2IO` read-only io, with functions to convert and export. Also the new entries were wrongly placed in 0.13.0...now moved to upcoming

src/hdmf_zarr/backend_zarrv2.py
alejoe91: done now :)

tests/unit/test_fsspec_streaming.py
alejoe91: Issue: https://github.com/hdmf-dev/hdmf-zarr/issues/357

tests/unit/helpers/generate_v2_nwb_zarr.py
```diff
@@ -0,0 +1,137 @@
...
+parser.add_argument(
+    "--nwb-output-path",
+    type=str,
+    help="Path to write the output NWB zarr file (e.g. /tmp/test_file.zarr)",
+)
+parser.add_argument(
+    "--expectations-output-path",
+    type=str,
+    help="Path to write the JSON file with metadata expectations (e.g. /tmp/expectations.json)",
+)
```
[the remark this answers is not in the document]
alejoe91: this is mainly internal, I think it's ok to leave it as is

alejoe91:
@oruebel anything else on my end? Should I generate a small zarrv2 file and add it to the repo for testing?

oruebel:
> Should I generate a small zarrv2 file and add it to the repo for testing?

Yes, please go ahead.

alejoe91:
Done. Added test file generated with pynwb 2.8.3 and Zarr v2.18

oruebel (approved):
Thanks @alejoe91 . From my end this looks good. @rly I think it would be good if you could also take a look since this is part of the larger Zarr V3 upgrade.

src/hdmf_zarr/backend_zarrv2.py
```diff
@@ -0,0 +1,646 @@
...
+
+    # ----- group iteration with v2 chunk fallback -----
+
+    def _iter_children(self, zarr_obj):
+        """Yield ``(name, child)`` pairs for the children of *zarr_obj*.
+
+        This overrides :meth:`ZarrIO._iter_children`. The base implementation
+        relies on ``zarr_obj.groups()`` / ``zarr_obj.arrays()``, but those calls
+        ask zarr-python v3 to open every child, which raises for zarr v2 arrays
+        that v3 cannot parse (most commonly object-dtype arrays stored with
```
d-v-b, line 472: > which raises for zarr v2 arrays
        that v3 cannot parse (most commonly object-dtype arrays stored with v2-only codecs such as ``pickle``, ``json2``, or ``vlen-utf8``)

vlen-utf8 should work, but more generally this is behavior I would like to fix in zarr-python. We should support gradual levels of support for arrays with unknown data types, so e.g. you could see that the array is there, but reading / writing chunks would be disallowed.

alejoe91: Thanks @d-v-b 

I think we should keep an eye on this and defer the logic to zarr-python once it's implemented. I opened an issue on our end as a reminder: #358

alejoe91:
@oruebel @rly can we move forward with this?

oruebel:
> @oruebel @rly can we move forward with this?

Looks fine from my end. I leave final review to @rly

rly:
Thanks. I'm starting to review this now. I made some minor fixes:

- I renamed the test file (combat → compat)
- I fixed broken Sphinx refs in `backend.py`, `nwb_zarrv2.py`, and `overview.rst`.
- I corrected the backward-compat test docstring.
- I moved the unreleased #348 and this PR CHANGELOG entries to Upcoming Release.
- I reflowed/formatted the new v2 backend modules with black

And I added a more informative error message if a user tries to read a Zarr v2 with the new `ZarrIO`/`NWBZarrIO` class:
> Failed to read '…' with NWBZarrIO, which reads Zarr v3 files, but this path is a Zarr v2 file. Open it read-only with the NWBZarrV2IO. Or convert it to Zarr v3 with NWBZarrV2IO.convert_to_v3(source_path, dest_path).

rly:
A Zarr to HDF5 test is timing out sporadically due to a tricky deadlocking issue between the HDF5 global lock, Python's cyclic garbage collector, and zarr v3's async→sync execution model. This should be resolved in HDMF: https://github.com/hdmf-dev/hdmf/pull/1547

rly:
I made a few more minor changes:

- Fixed the "this is a Zarr v2 file, use `NWBZarrV2IO`" hint, which wasn't showing up when the error came from `read_builder`.
- Made `ZarrV2IO` fall back to a non-consolidated open more reliably, and removed an unused import.
- Cleaned up the changelog (moved the #348 entry to 0.13.0, added one for the `zarr.Array.__iter__` patch).
- Small docstring/comment clarifications (s3 anonymous reads, the 1-D object-array assumption, link to #356).
- Ran black on the files flagged by the pinned hook.

tests/unit/test_v2_backward_compat.py
rly: I think 0.14 is fine. 1.0 is scary. Let's keep it pre-1.0 until we have some more real tests.

rly (approved):
LGTM. Thank you so much @alejoe91 !

PR #360, opened 2026-07-22
[DIAGNOSTIC] Instrument the intermittent Python 3.14 CI hang

rly:
## Motivation

The Python 3.14 test jobs intermittently **deadlock** and, because the jobs have no `timeout-minutes`, run to GitHub's **6-hour** cap before being cancelled (e.g. run [#1372](https://github.com/hdmf-dev/hdmf-zarr/actions/runs/29768795216)).

The hang is **flaky and not tied to this or any single change**:
- Across runs it has hit *different* jobs, both optional and non-optional, on linux, windows, and conda.
- It also occurred on the `zarr-v3-migration` base branch (run #1369, July 7), where the `-optional` jobs hung while non-optional passed, the opposite of the most recent run.
- Other runs of the same branches passed cleanly.
- It does not reproduce in a local `python:3.14` container across the optional/non-optional/scipy/hdf5plugin permutations, consistent with a **timing-dependent threading / event-loop deadlock** on the multi-CPU CI runners (prime suspects: zarr 3's `sync()` background loop, or the multiprocessing in `test_parallel_write.py`).

There is currently **no visibility**: a hung job produces no stack trace, so the root cause can't be diagnosed.

## What this draft adds

- **`pytest-timeout`** (in the `test` dependency group) plus a global `[tool.pytest.ini_options]` timeout of **300s** with `timeout_method = "thread"`. On a hang this dumps **every thread's stack** and kills the process. The thread method is cross-platform and works even when the main thread is stuck in a C call (e.g. HDF5), which the signal method cannot interrupt. Verified locally: a hung test now prints `Timeout / Stack of MainThread ...` and fails fast.
- **`timeout-minutes: 45`** on the `run_tests.yml` jobs, a backstop for hangs *outside* a test (e.g. during install/tox setup), so a job fails in minutes rather than 6 hours.
- **`diagnose_py314_hang.yml`**, a dedicated workflow that runs the unit suite in a **loop** across the suspect py3.14 configs (linux / conda-linux / windows, optional and non-optional) to trip the flake within a single run. Re-run it from the Actions tab (`workflow_dispatch`) with an adjustable iteration count to keep rolling the dice; a caught hang leaves an actionable traceback in the log.

## How to use

1. Let this PR's CI run. If a normal job hangs, it will now self-terminate at the 300s per-test timeout with a thread dump instead of spinning.
2. To actively hunt it, trigger **Diagnose py3.14 CI hang** from the Actions tab (bump `iterations`, e.g. 20). Re-run until a job catches the deadlock, then read the dumped stacks to find where it's stuck.
3. Once the culprit is identified, the fix lands separately; the `timeout-minutes` and `pytest-timeout` bits are worth keeping regardless.

Draft: this is instrumentation to find the root cause, not the fix itself.

### DIFF
--- a/.github/workflows/diagnose_py314_hang.yml
+++ b/.github/workflows/diagnose_py314_hang.yml
@@ -12,7 +12,7 @@ on:
     inputs:
       iterations:
         description: "Times to run the unit test suite per job"
-        default: "10"
+        default: "40"
   pull_request:
     paths:
       - ".github/workflows/diagnose_py314_hang.yml"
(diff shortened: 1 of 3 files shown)


rly:
Superseded by #361. #360 is based on `dev`, which is still zarr v2 (synchronous, no async `sync()` loop), so the py3.14 deadlock cannot reproduce here. The diagnostic + focused hunt now live on #361, based on `zarr-v3-migration` where the hang actually occurs.

PR #363, opened 2026-07-23
Fix Zarr-to-HDF5 export deadlock

### LINKED ISSUE #362: Deadlock exporting a Zarr v3 file to HDF5
## Summary

Exporting an hdmf-zarr (Zarr v3) file to HDF5 (`HDF5IO.export(src_io=ZarrIO(...))`) can **deadlock permanently**. It is intermittent: a single export usually succeeds, but under repetition / concurrency it hangs reliably. This is the cause of the 6-hour Python 3.14 CI hangs in the `ZarrToHDF5` `DynamicTable` export-roundtrip tests in the Zarr v3 support branches.

It is a three-way interaction between the HDF5 global lock, Python's cyclic garbage collector, and zarr v3's async→sync execution model — and it does **not** occur with Zarr v2.

## Reproducer (hdmf-zarr)

Runs `ZarrIO -> HDF5IO` export in a loop across processes. Based on `tests/unit/test_io_convert.py::TestZarrToHDF5DynamicTableC0::test_export_roundtrip`

```python
import multiprocessing as mp, time, os, tempfile
from hdmf.common import DynamicTable, get_manager
from hdmf.backends.hdf5 import HDF5IO
from hdmf_zarr import ZarrIO
from hdmf_zarr.backend import ROOT_NAME
from zarr.storage import LocalStore

def make_table():
    t = DynamicTable(name=ROOT_NAME, description="an example table")
    t.add_column("foo", "an int column"); t.add_column("bar", "a float column")
    t.add_column("qux", "a boolean column"); t.add_column("quux", "a enum column", enum=True, index=False)
    t.add_row(foo=27, bar=28.0, qux=True, quux="a"); t.add_row(foo=37, bar=38.0, qux=False, quux="b")
    return t

def work(beat):
    d = tempfile.mkdtemp(); hpath = os.path.join(d, "export.h5")
    # one path passed as a str, one as a LocalStore, to exercise both store forms
    write_paths = [os.path.join(d, "w.zarr"), LocalStore(os.path.join(d, "w2.zarr"))]
    while True:
        ios = []
        for wp in write_paths:
            with ZarrIO(wp, manager=get_manager(), mode="w") as wio:
                wio.write(make_table())
            with ZarrIO(wp, manager=get_manager(), mode="r") as rio:
                with HDF5IO(hpath, manager=get_manager(), mode="w") as eio:
                    eio.export(src_io=rio, write_args={"link_data": False})
            back = HDF5IO(hpath, manager=get_manager(), mode="r"); ios.append(back); back.read()
        for io in ios: io.close()
        beat.value = time.time()

if __name__ == "__main__":
    beats = [mp.Value("d", time.time()) for _ in range(10)]
    procs = [mp.Process(target=work, args=(b,), daemon=True) for b in beats]
    for p in procs: p.start()
    time.sleep(120)
    stalled = sum(time.time() - b.value > 40 for b in beats)
    print(f"{stalled}/10 processes DEADLOCKED (no progress >40s)" if stalled else "no deadlock")
    for p in procs: p.terminate()
```

**Result:** `10/10 processes DEADLOCKED`. (Reducing to a single small export usually succeeds; the loop + read-back + two write paths generate the finalizable IO cycles that make the race reliable.)

## Root cause

Three threads participate in a lock-ordering deadlock (from the captured `faulthandler` dump):

- **Main thread** — inside an h5py write of a zarr-backed dataset, so it holds the process-wide HDF5 lock (`phil`, h5py's reentrant module-level lock), and is blocked in zarr's sync bridge:
  ```
  h5py Dataset.__setitem__            # dset[:] = <zarr array>, holds 'phil'
    numpy.asarray(zarr_array)
    zarr Array.__getitem__ -> get_basic_selection -> zarr.core.sync.sync()   # waits on the read
  ```
  h5py materializes the zarr array lazily via `numpy.asarray()` *inside* `__setitem__`, so `sync()` is called **while `phil` is held**.

- **`zarr_io` thread** (zarr v3's single background event loop) — running the read, but a **cyclic GC pass fired on this thread** and is executing a finalizer:
  ```
  run_forever -> _run_once -> (zarr read coroutine, mid _get_selection)
    HDMFIO.__del__ (hdmf/backends/io.py) -> HDF5IO.close() -> h5py File.close()   # wants 'phil'
  ```
  `HDMFIO.__del__` does a **blocking `h5py` close**, which needs `phil`.

- **executor workers** — idle.

So: the main thread holds `phil` and waits on zarr's event loop; the event-loop thread is running a GC-triggered `HDMFIO.__del__` that blocks on `phil`. Neither proceeds — permanent deadlock.

The finalizer lands on the event-loop thread because hdmf IO objects live in reference cycles (io ↔ manager ↔ builders): they are reaped by the **cyclic GC**, which runs on whichever thread trips its threshold — here, zarr's event-loop thread mid-read.

<details>

<summary>Minimal demonstration without hdmf</summary>

## Minimal demonstration without hdmf

The same deadlock with only `zarr`, `h5py`, and `numpy`. `IOish` stands in for an `HDMFIO` whose `__del__` closes an HDF5 file; the self-reference makes it *cyclic* garbage so the cyclic GC (not refcounting) reaps it — potentially on zarr's loop thread.

```python
import multiprocessing as mp, time, os, tempfile
import numpy as np, zarr, h5py

class IOish:
    def __init__(self):
        self._f = h5py.File(f"m{id(self)}", "w", driver="core", backing_store=False)
        self._cycle = self  # cyclic ref -> reaped by the cyclic GC (runs on any thread)
    def __del__(self):
        self._f.close()     # blocking h5py close -> wants the HDF5 lock

def work(beat):
    z = zarr.create_array(store={}, shape=(20000,), chunks=(8,), dtype="f8")
    z[:] = np.arange(20000)
    out = os.path.join(tempfile.mkdtemp(), "o.h5")
    while True:
        for _ in range(5):
            IOish()  # leak finalizable h5py-closers into cyclic garbage
        with h5py.File(out, "w") as f:
            d = f.create_dataset("x", shape=z.shape, dtype=z.dtype)
            d[:] = z  # holds the HDF5 lock; asarray(z) -> zarr sync() read
        beat.value = time.time()

if __name__ == "__main__":
    beats = [mp.Value("d", time.time()) for _ in range(10)]
    procs = [mp.Process(target=work, args=(b,), daemon=True) for b in beats]
    for p in procs: p.start()
    time.sleep(60)
    stalled = sum(time.time() - b.value > 20 for b in beats)
    print(f"{stalled}/10 processes DEADLOCKED (no progress >20s)" if stalled else "no deadlock")
    for p in procs: p.terminate()
```

**Result:** `10/10 processes DEADLOCKED` (within ~26s). Two controls confirm the mechanism:
- Remove `self._cycle` (so the finalizer runs immediately by refcounting on the creating thread) → **`no deadlock`**.
- Install Zarr v2 instead (adapt `zarr.create_array` → `zarr.zeros`) → **`no deadlock`** (see below).

</details>

## Zarr v2 vs v3

Zarr v2 is synchronous: reads run on the calling thread, with no `sync()` and no background event-loop thread. If the cyclic GC fires mid-read there, the finalizer runs on the *same* thread that holds `phil`, and `phil` is reentrant, so it just succeeds. Verified: the v2 form of the minimal example reports `no deadlock`.

Zarr v3 dispatches reads to a separate `zarr_io` thread via `sync()`, which is what puts the finalizer on a *different* thread from the `phil` holder.

## Suggested fixes (most targeted first)

1. **hdmf-zarr export path** — don't let h5py pull data lazily through `numpy.asarray()` while `phil` is held: materialize each source array to numpy *before* entering the h5py write. This closes the window where `sync()` is called under `phil`.
   - **RAM note:** this does **not** increase peak memory. `dset[:] = <zarr array>` already calls `numpy.asarray()` on the whole array inside `__setitem__`, so the full array is materialized in RAM today. Materializing one line earlier has the same peak.
2. **hdmf** — `HDMFIO.__del__` performing a blocking `close()` is the aggravating factor; a finalizer that can run on any thread should avoid blocking I/O.
3. **zarr** — the enabler is that `sync()` runs the read on a separate event-loop thread, so a Python finalizer that fires there (via cyclic GC) executes off the thread holding `phil`. This is **not** addressed by the codec-pipeline rework (zarr-developers/zarr-python#3720, #3715, #3885): the deadlock still reproduces on current `main` with both the default `BatchedCodecPipeline` and the opt-in `FusedCodecPipeline`, since both still dispatch reads through `sync()`.

## Environment

Confirmed on h5py 3.16.0, hdmf 6.1.0, numpy 2.5.1, Python 3.14.6 (Linux). Reproduces on zarr **3.2.1** (released) and zarr **main** (3.2.2.dev) — with both the default `BatchedCodecPipeline` and the opt-in `FusedCodecPipeline`.

Not Python-3.14-specific: the minimal reproducer above also gives `10/10 processes DEADLOCKED` on a standard GIL-enabled **Python 3.12.13** (macOS, zarr 3.2.1, h5py 3.16.0, numpy 2.5.1). The blocking h5py close releases the GIL at the C level, so `phil`, not the GIL, is the lock that never gets handed back. The deadlock also reproduces in both the standard GIL and no-GIL 3.14 builds.

---

*This report was drafted and confirmed with assistance from AI (Claude Opus 4.8).*

rly:
Fixes #362.

Patches `zarr.Array.__array__` so numpy coercion of a zarr Array reads on the calling thread. During `HDF5IO.export`, h5py copies the array while holding its global `phil` lock; zarr's default read blocks on the shared `zarr_io` loop thread, and a cyclic-GC finalizer that acquires `phil` on that thread deadlocks against the caller. Reading on the calling thread keeps the loop thread out of the read, so the lock cycle cannot form.

Verified: an end-to-end ZarrIO→HDF5 export reproducer deadlocks 10/10 processes without this change and 0/10 with it. Adds a deterministic (non-racy) regression test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 - Fixed bug where `ZarrIO.generate_dataset_html` would raise an error when called with a non-Zarr object. @oruebel [#355](https://github.com/hdmf-dev/hdmf-zarr/pull/355)
+- Fixed a deadlock when exporting a Zarr file to HDF5 with `HDF5IO.export`. numpy coercion of a zarr Array now reads on the calling thread instead of zarr's shared background event-loop thread, so a garbage-collection finalizer that acquires h5py's global lock on that thread cannot deadlock against the export. @rly [#362](https://github.com/hdmf-dev/hdmf-zarr/issues/362)
 
 
 ## 0.13.0 (June 22, 2026)
--- a/src/hdmf_zarr/backend.py
+++ b/src/hdmf_zarr/backend.py
@@ -1,6 +1,7 @@
 """Module with the Zarr-based I/O-backend for HDMF"""
 
 # Python imports
+import asyncio
 import itertools
 import json
 import os
@@ -54,6 +55,39 @@ def _zarr_array_getitem_scalar_fix(self, key):
 
 
 Array.__getitem__ = _zarr_array_getitem_scalar_fix
+
+
+# zarr v3 executes reads on a shared background event-loop thread named "zarr_io":
+# a synchronous read blocks the calling thread while that thread runs the read.
+# numpy calls Array.__array__ to coerce a zarr Array into an ndarray, which happens
+# when another backend copies the array, e.g. h5py during export of a Zarr file to
+# HDF5. h5py holds its global "phil" lock across that copy, and h5py's own object
+# finalizers also acquire "phil". If a cyclic-GC finalizer runs on the "zarr_io"
+# thread while the calling thread holds "phil" and waits on that thread's read, the
+# two threads deadlock on "phil". Materializing the array on the calling thread keeps
+# the "zarr_io" thread out of the read, so the lock cycle cannot form.
+_zarr_array_original_array = Array.__array__
+
+
+def _zarr_array_array_on_calling_thread(self, dtype=None, copy=None):
+    if copy is False:
+        raise ValueError("`copy=False` is not supported. This method always creates a copy.")
+    try:
+        asyncio.get_running_loop()
+    except RuntimeError:
+        # No event loop runs on this thread, so read the whole array here on a private
+        # loop, which keeps the read on the calling thread.
+        loop = asyncio.new_event_loop()
+        try:
+            data = loop.run_until_complete(self._async_array.getitem(Ellipsis))
+        finally:
+            loop.close()
+        return np.asarray(data, dtype=dtype)
+    # An event loop is already running on this thread; use zarr's default coercion.
+    return _zarr_array_original_array(self, dtype=dtype, copy=copy)
+
+
+Array.__array__ = _zarr_array_array_on_calling_thread
 
 
 # Module variables
--- a/tests/unit/test_array_materialization.py
+++ b/tests/unit/test_array_materialization.py
@@ -1,0 +1,72 @@
+"""Tests for reading a full zarr Array to numpy on the calling thread.
+
+hdmf_zarr.backend patches ``zarr.Array.__array__`` so that numpy coercion of a zarr
+Array (e.g. when h5py copies it during a Zarr -> HDF5 export) reads on the calling
+thread. This keeps zarr's shared "zarr_io" event-loop thread free of the read, so a
+cyclic-GC finalizer that acquires h5py's "phil" lock on that thread cannot deadlock
+against the caller holding "phil".
+"""
+
+import asyncio
+import time
+from unittest import TestCase
+
+import numpy as np
+import zarr
+from zarr.core.sync import _get_loop
+
+# Importing the backend installs the Array.__array__ patch.
+import hdmf_zarr.backend  # noqa: F401
+
+
+class TestArrayMaterialization(TestCase):
+
+    def test_asarray_values_1d(self):
+        z = zarr.create_array(store={}, shape=(50,), chunks=(8,), dtype="f8")
+        z[:] = np.arange(50)
+        np.testing.assert_array_equal(np.asarray(z), np.arange(50))
+
+    def test_asarray_values_2d(self):
+        z = zarr.create_array(store={}, shape=(10, 3), chunks=(4, 2), dtype="i4")
+        expected = np.arange(30).reshape(10, 3)
+        z[:] = expected
+        np.testing.assert_array_equal(np.asarray(z), expected)
+
+    def test_asarray_dtype_coercion(self):
+        z = zarr.create_array(store={}, shape=(5,), chunks=(2,), dtype="i4")
+        z[:] = np.arange(5)
+        arr = np.asarray(z, dtype="f4")
+        self.assertEqual(arr.dtype, np.dtype("f4"))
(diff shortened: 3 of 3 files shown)


rly:
Superseding this in favor of hdmf-dev/hdmf#1547, which fixes the deadlock in the right place.

This PR patched `zarr.Array.__array__` process-wide and materialized *inside* the window where h5py holds `phil` (its global lock), which forced a fresh-per-call event loop as a workaround (untested against aiohttp-backed remote stores). hdmf#1547 instead materializes the zarr array in `HDF5IO.__list_fill__`/`__scalar_fill__` *before* the write, removing the only window where `phil` is held across a zarr `sync()`. That uses zarr's normal, tested read path, is scoped to the export/write path (no process-wide behavior change), and is RAM-neutral. hdmf#1547 also makes `HDMFIO.__del__` non-blocking (ResourceWarning + atexit) to remove the unsafe blocking-close-in-a-finalizer for the whole IO class.

Closing here; the fix will land via the next HDMF release (bump the `hdmf>=` pin then).

PR #365, opened 2026-08-25
Resolve compound fields declared `uint` instead of writing them as float64

### LINKED ISSUE #364: [Bug]: Compound fields with spec dtype `uint` are written as float64 (breaks HERD round-trip)
### What happened?

`ZarrIO.__dtypes` has no entry for the bare spec type `uint`. Because the lookup uses
`dict.get`, an unresolved spec type silently becomes `None`, and inside the compound-dtype
branch that `None` is handed straight to `np.dtype([...])`, where numpy reads it as
"unspecified" and substitutes `float64`.

The result is that every compound field declared as `dtype: uint` is written as `float64`.
In `hdmf-common-schema` that is all five HERD index fields (`objects.files_idx`,
`object_keys.{objects_idx,keys_idx}`, `entity_keys.{entities_idx,keys_idx}`), so a HERD
round-tripped through Zarr comes back with float indices and the row-based accessors raise.

I expected the index fields to round-trip as unsigned integers, as they do on HDF5.

**Root cause**

`ZarrIO.__dtypes` ([backend.py L1362](https://github.com/hdmf-dev/hdmf-zarr/blob/5e633ffd600ae5b5468fd1812d30b470b8ae6d31/src/hdmf_zarr/backend.py#L1362))
has `uint8`/`uint16`/`uint32`/`uint64` but no bare `uint`, while `HDF5IO.__dtypes` does. It has
bare `int` -> `int32` and bare `float` -> `float32`, so this looks like an omission rather than a
deliberate choice:

```
zarr  __dtypes['uint'] -> None
hdf5  __dtypes['uint'] -> <class 'numpy.uint32'>

uint keys in zarr table: ['uint16', 'uint32', 'uint64', 'uint8']
uint keys in hdf5 table: ['uint', 'uint16', 'uint32', 'uint64', 'uint8']
```

The miss is silent for most datasets, because `__resolve_dtype__` falls back to inferring from
the data when the helper returns `None` ([L1413-L1417](https://github.com/hdmf-dev/hdmf-zarr/blob/5e633ffd600ae5b5468fd1812d30b470b8ae6d31/src/hdmf_zarr/backend.py#L1413-L1417)):

```python
@classmethod
def __resolve_dtype__(cls, dtype, data):
    dtype = cls.__resolve_dtype_helper__(dtype)
    if dtype is None:
        dtype = cls.get_type(data)      # <- rescues the plain-dataset case
    return dtype
```

But compound dtypes are resolved by a recursive branch *inside* the helper, which builds the
numpy dtype before returning, so the fallback never sees the `None` ([L1429-L1430](https://github.com/hdmf-dev/hdmf-zarr/blob/5e633ffd600ae5b5468fd1812d30b470b8ae6d31/src/hdmf_zarr/backend.py#L1429-L1430)):

```python
elif isinstance(dtype, list):
    return np.dtype([(x["name"], cls.__resolve_dtype_helper__(x["dtype"])) for x in dtype])
```

```python
>>> np.dtype([('files_idx', None)])
dtype([('files_idx', '<f8')])
```

So the two backends resolve HERD's `objects` spec differently:

```
ZarrIO  -> [('files_idx', '<f8'), ('object_id', '<U')]
HDF5IO  -> [('files_idx', '<u4'), ('object_id', 'O')]
```

**Notes**

- Write and read both succeed; only construction of a `Row` fails, when docval type-checks the
  field. `HERD.to_dataframe()` works fine, so this can go unnoticed until someone calls
  `get_object_entities` or anything else that goes through `_find_object`.
- Not specific to the Zarr v3 migration. hdmf-zarr 0.13.0 has the same missing key; there the
  compound dataset goes through the zarr-2 object-codec path and comes back as `object` dtype,
  so it fails at a different line (`TypeError: 'int' object is not subscriptable`) from the same
  cause. I hit this on the `zarr-v3-migration` branch (#325) but it is pre-existing.
- Possibly related to #323, which asks about `uint` handling for attributes; that is a different
  mechanism (JSON attributes) but the same spec type.

**Suggested fix**

Add `"uint": np.uint32` to `ZarrIO.__dtypes` to match `HDF5IO`. It may also be worth making the
`isinstance(dtype, list)` branch raise on a field dtype it cannot resolve instead of silently
emitting `None`, so that any future missing key surfaces as an error rather than as quietly
corrupted data.

I am happy to open a PR for this if it would help.

### Steps to Reproduce

```shell
import datetime
from uuid import uuid4

import zarr
from pynwb import NWBFile
from pynwb.file import Subject

nwbfile = NWBFile(
    session_description="HERD compound dtype repro",
    identifier=str(uuid4()),
    session_start_time=datetime.datetime.now(datetime.timezone.utc),
)
nwbfile.subject = Subject(subject_id="s1", species="Mus musculus")

herd = nwbfile.get_external_resources()
herd.add_ref(
    container=nwbfile.subject,
    key=nwbfile.subject.species,
    entity_id="NCBITaxon:10090",
    entity_uri="http://purl.obolibrary.org/obo/NCBITaxon_10090",
)

path = "herd_repro.nwb.zarr"
with NWBZarrIO(path, mode="w") as io:
    io.write(nwbfile)

# The index fields were written as float64 rather than uint.
objects = zarr.open_array(f"{path}/general/external_resources/objects", mode="r")
print("objects dtype:", objects.dtype)

with NWBZarrIO(path, mode="r") as io:
    read_nwbfile = io.read()
    read_herd = read_nwbfile.external_resources
    print("to_dataframe() rows:", len(read_herd.to_dataframe()))  # works
    read_herd.get_object_entities(container=read_nwbfile.subject)  # raises

# Prints: objects dtype: [('files_idx', '<f8'), ('object_id', '<U512'), ...]
# then raises (traceback below). Swapping NWBZarrIO for NWBHDF5IO runs to completion and
# writes the same dataset as [('files_idx', '<u4'), ('object_id', 'O'), ...].
```

### Traceback

```shell
  File "herd_repro.py", line 36, in <module>
    read_herd.get_object_entities(container=read_nwbfile.subject)  # raises
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../hdmf/utils.py", line 641, in func_call
    return func(args[0], **pargs)
           ^^^^^^^^^^^^^^^^^^^^^^
  File ".../hdmf/common/resources.py", line 841, in get_object_entities
    object_field = self._find_object(file, target_container, target_relative_path, field)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../hdmf/common/resources.py", line 340, in _find_object
    return self.objects.row[matches[0]]
           ~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File ".../hdmf/container.py", line 1569, in __getitem__
    ret = self.table.__rowclass__(*row, table=self.table, idx=idx)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../hdmf/utils.py", line 640, in func_call
    pargs = _check_args(args, kwargs)
            ^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../hdmf/utils.py", line 633, in _check_args
    raise ExceptionType(msg)
TypeError: Row.__build_row_class.<locals>.__init__: incorrect type for 'files_idx' (got 'float64', expected 'int or integer')
```

### Operating System

macOS

### Python Version

3.11

### Package Versions

Reproduced on the `zarr-v3-migration` branch (#325):

```
python     3.11.7
hdmf       6.2.0
hdmf_zarr  0.13.1.dev27+g003a83e78
numcodecs  0.16.5
numpy      2.4.6
pynwb      4.1.0
zarr       3.1.6
```

The missing `uint` key is also present on the current release, where the same HERD round-trip
fails with a different symptom:

```
hdmf-zarr 0.13.0 | zarr 2.18.7 | hdmf 6.1.0 | pynwb 4.1.0
```

---

*Edit: corrected the `__dtypes` line reference (it pointed at the line number on the `zarr-v3-migration` branch, not `dev`) and pinned the source links to `5e633ff`.*

ehennestad:
Fixes #364.

## The bug

`ZarrIO.__dtypes` has no entry for the bare spec type `uint`. The lookup uses `dict.get`, so an
unknown spec type resolves to `None` rather than raising.

For a plain dataset that is harmless, because `__resolve_dtype__` recovers:

```python
@classmethod
def __resolve_dtype__(cls, dtype, data):
    dtype = cls.__resolve_dtype_helper__(dtype)
    if dtype is None:
        dtype = cls.get_type(data)      # <- infers from the data
    return dtype
```

But a compound dtype is built inside `__resolve_dtype_helper__` itself, so the `None` reaches
`np.dtype` before the fallback can see it, and numpy reads it as "unspecified":

```python
>>> np.dtype([('files_idx', None)])
dtype([('files_idx', '<f8')])
```

Every compound field declared `dtype: uint` was therefore written as `float64`. In
hdmf-common-schema that is all five HERD index fields (`objects.files_idx`,
`object_keys.{objects_idx,keys_idx}`, `entity_keys.{entities_idx,keys_idx}`), so a HERD read back
from Zarr has float indices and the row lookup rejects them:

```
TypeError: incorrect type for 'files_idx' (got 'float64', expected 'int or integer')
```

Write and read both succeed, and `HERD.to_dataframe()` works, so this can go unnoticed until
something calls `get_object_entities` or anything else routed through `_find_object`.

## The change

- Add the missing `uint` -> `np.uint32` key, matching `HDF5IO`. `short` -> `np.int16` was missing
  for the same reason and is added alongside it; those were the only two spec aliases present in
  `HDF5IO.__dtypes` and absent here that map cleanly onto Zarr. (`datetime` is also absent, but
  `HDF5IO` maps it to `object`, which means something different in this table, so I left it alone.)
- Raise on a compound field dtype that cannot be resolved, instead of passing `None` to `np.dtype`.
  Without this, the next missing key would again produce silently corrupted data rather than an
  error. The scalar fallback is deliberately untouched and is covered by a test.

Both are small enough to split if you would rather take only the table entry.

## Verification

Before and after, writing a compound dataset whose spec declares `{"name": "idx", "dtype": "uint"}`:

```
before -> [('idx', '<f8'), ('value', '<f4')]
after  -> [('idx', '<u4'), ('value', '<f4')]
```

Six tests are added in `TestResolveCompoundDtype`. Five of them fail on `dev` without the change;
the sixth guards the scalar fallback and passes either way, which is the point of it.

The full unit suite passes on this branch (551 passed, 13 skipped).

## Scope

This fixes dtype **resolution** on `dev`, which is where the bug originates, but it does not by
itself fix the HERD round-trip here: on zarr-python 2 a compound dtype containing strings goes
through the object-codec path, so `objects` comes back as `object` dtype and fails for a separate
reason. On the `zarr-v3-migration` branch (#325), where the resolved compound dtype is honoured,
this change fixes the HERD round-trip end to end — I verified that `get_object_entities` returns
the annotation correctly with the patch applied there.

I based the PR on `dev` since that is where the missing key lives; it should carry over to #325
unchanged. Happy to retarget if you would rather it land there first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 - Fixed bug where `ZarrIO.generate_dataset_html` would raise an error when called with a non-Zarr object. @oruebel [#355](https://github.com/hdmf-dev/hdmf-zarr/pull/355)
-- Fixed bug where a compound dtype field declared with the spec type `uint` (or `short`) was written as `float64`, because `ZarrIO.__dtypes` was missing those keys and the resulting `None` was passed to `np.dtype`. This affected the `HERD` index fields, among others. @ehennestad [#PRNUM](https://github.com/hdmf-dev/hdmf-zarr/pull/PRNUM)
+- Fixed bug where a compound dtype field declared with the spec type `uint` (or `short`) was written as `float64`, because `ZarrIO.__dtypes` was missing those keys and the resulting `None` was passed to `np.dtype`. This affected the `HERD` index fields, among others. @ehennestad [#365](https://github.com/hdmf-dev/hdmf-zarr/pull/365)
 
 
 ## 0.13.0 (June 22, 2026)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 - Fixed bug where `ZarrIO.generate_dataset_html` would raise an error when called with a non-Zarr object. @oruebel [#355](https://github.com/hdmf-dev/hdmf-zarr/pull/355)
+- Fixed bug where a compound dtype field declared with the spec type `uint` (or `short`) was written as `float64`, because `ZarrIO.__dtypes` was missing those keys and the resulting `None` was passed to `np.dtype`. This affected the `HERD` index fields, among others. @ehennestad [#PRNUM](https://github.com/hdmf-dev/hdmf-zarr/pull/PRNUM)
 
 
 ## 0.13.0 (June 22, 2026)
--- a/src/hdmf_zarr/backend.py
+++ b/src/hdmf_zarr/backend.py
@@ -1369,8 +1369,10 @@ class ZarrIO(HDMFIO):
         "uint64": np.uint64,
         "int": np.int32,
         "int32": np.int32,
+        "short": np.int16,
         "int16": np.int16,
         "int8": np.int8,
+        "uint": np.uint32,
         "bool": np.bool_,
         "bool_": np.bool_,
         "text": str,
@@ -1427,7 +1429,19 @@ class ZarrIO(HDMFIO):
         elif isinstance(dtype, dict):
             return cls.__dtypes.get(dtype["reftype"])
         elif isinstance(dtype, list):
-            return np.dtype([(x["name"], cls.__resolve_dtype_helper__(x["dtype"])) for x in dtype])
+            # A field dtype that does not resolve must not be passed on to np.dtype, which reads
+            # None as "unspecified" and silently substitutes float64. Unlike the scalar case,
+            # __resolve_dtype__ cannot fall back to inferring the dtype from the data here,
+            # because the None is consumed while the compound dtype is being built.
+            fields = []
+            for field in dtype:
+                field_dtype = cls.__resolve_dtype_helper__(field["dtype"])
+                if field_dtype is None:
+                    raise ValueError(
+                        f"Can't resolve dtype {field['dtype']!r} for compound field {field['name']!r}"
+                    )
+                fields.append((field["name"], field_dtype))
+            return np.dtype(fields)
         else:
             raise ValueError(f"Can't resolve dtype {dtype}")
 
--- a/tests/unit/test_zarrio.py
+++ b/tests/unit/test_zarrio.py
@@ -25,6 +25,8 @@ import numpy as np
 from hdmf_zarr.backend import ZarrIO, ROOT_NAME
 from .helpers.utils import BuildDatasetShapeMixin, BarData, BarDataHolder
 from hdmf.spec import DatasetSpec
+from hdmf.build import GroupBuilder, DatasetBuilder
+from hdmf.backends.hdf5.h5tools import HDF5IO
 import os
 import shutil
 import warnings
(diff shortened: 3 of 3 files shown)


ehennestad:
@bendichter

oruebel:
@ehennestad thanks for the PR. This looks good to me. I think it may be useful to update this to target the `zarr-v3-migration](https://github.com/hdmf-dev/hdmf-zarr/tree/zarr-v3-migration) branch instead to make sure this is being fixed going forward since that PR introduces separate V2 and V3 backends. What do you think?

oruebel (approved):

PR #333, opened 2026-03-03
Remove zarr.Array monkey-patches, use ndim-based detection

h-mayorquin:
## Summary

Testing the approach proposed in https://github.com/hdmf-dev/hdmf-zarr/pull/325#discussion_r2876481460.

- Merges dev into the zarr-v3-migration branch and resolves conflicts
- Removes both `zarr.Array` monkey-patches (`__len__` and `__getitem__`)
- Replaces `hasattr(data, "__len__")` checks with `ndim`-based collection detection (`_is_collection`)
- Handles 0-d ndarrays in `get_type()` via `data.item()` extraction (`_resolve_scalar` pattern)

All 397 tests pass locally.

Related:
- zarr-python: https://github.com/zarr-developers/zarr-python/issues/3740, https://github.com/zarr-developers/zarr-python/issues/3741
- hdmf: https://github.com/hdmf-dev/hdmf/pull/1414, https://github.com/hdmf-dev/hdmf/pull/1415

## Test plan

- [x] Run `pytest tests/unit/test_zarrio.py` — 397 passed, 0 failed
- [ ] CI passes

PR #366, opened 2026-08-31
Propose changes to v3

oruebel:
## Motivation

Propose enhancements to the ZarrV3 migrations

## Proposed changes

- https://github.com/hdmf-dev/hdmf-zarr/pull/333/ proposed to replace monkey patching Zarr Array. Within the ZarrV3 backend those changes help avoid relying on the monkey patches. However, with the latest version of HDMF and PyNWB this approach leads to test failures because HDMF/PyNWB: 1) use `len` on arrays, e.g., for validation (e.g, `len(args_to_set['electrodes'].data)` in `ElectricalSeries`, 2) use `isinstance(data, collections.abc.Iterable)` to check for iterable which fail `docval` without `__iter__`, 3) scalar indexing returns a 0-d array instead of a python scalar. [PyNWB#2235](https://github.com/NeurodataWithoutBorders/pynwb/pull/2235) audits PyNWB to remove reliance on (most of) these concepts of an array (so waiting on PyNWB should resolve this issue as well). However, to move this forward and to support lazy read of strings (see next item) we address this here. To avoid monkey-patching Zarr Array directly, this PR creates a new class `HDMFZarrArray(Array)` which inherits from `Zarr.Array` and adds the missing methods (`__len__`, `__iter___`, `__getitem__`). On read, we then patch the class of the instances that are being read. 
- With the `HDMFZarrArray` approach above we can also achieve lazy reading of String arrays. The approach discussed in https://github.com/hdmf-dev/hdmf-zarr/pull/325#discussion_r3515151691 was reverted because the wrapper class for string arrays did not pass HDMF validation. However, with `HDMFZarrArray` we are providing a Zarr array that passes validation and adds custom functionality, so we can extend `HDMFZarrArray` to handle the coding of strings on read. Implemented in [5d6ed2d](https://github.com/hdmf-dev/hdmf-zarr/pull/366/commits/5d6ed2d7ba849f811b0384502f20575978d3803d)
-  Fixed untested, broken code path where serializing a 0-D NumPy array attribute used  `value.item`  instead of  `value.item()` , leaving a method object that Zarr could not JSON-serialize and scalar byte arrays also were not decoded after unwrapping. See [488570d](https://github.com/hdmf-dev/hdmf-zarr/pull/366/commits/488570df01b6ccf88ff906410a3ad4a179b1a9b2)
- Addressed review  [r3908606927](https://github.com/hdmf-dev/hdmf-zarr/pull/325#discussion_r3908606927).  `from_h5py_dataset`  now lets explicit  compressor  and  filters  override inferred HDF5 codecs without duplicate-keyword errors.
- Addressed review [r3908606834](https://github.com/hdmf-dev/hdmf-zarr/pull/325#discussion_r3908606834) updated `ZarrV2IO._decode_v2_chunk` such that decoded object chunks now always reshape to the chunk’s N-D shape before assignment, avoiding divergent behavior for object and non-object chunks. See [8a760bd](https://github.com/hdmf-dev/hdmf-zarr/pull/366/commits/8a760bd0f2c3c8695b50624c716d8fa55b5f3c59) 
- Avoid unhandled Zarr async parse errors in `ZarrV2IO._iter_children`. Some v2 array metadata cannot be parsed by Zarr v3. Calling  `zarr_obj[entry]`  raises an exception that is caught for raw fallback, but leaves an internal async task whose exception appears after pytest completes. The fix validates raw  `.zarray`  metadata first and routes unsupported arrays directly to the raw decoder, while retaining normal lazy Zarr reads for compatible arrays. See [8a760bd](https://github.com/hdmf-dev/hdmf-zarr/pull/366/commits/8a760bd0f2c3c8695b50624c716d8fa55b5f3c59)
- Addressed review [r3908606781](https://github.com/hdmf-dev/hdmf-zarr/pull/325#discussion_r3908606781). Fixed v2 remote URL detection.  `is_zarr_v2_file`  now recognizes any protocol URL via  "://" , including GCS, Azure, and chained fsspec URLs, and opens it rather than incorrectly checking it as a local path. See [af78381](https://github.com/hdmf-dev/hdmf-zarr/pull/366/commits/af783814cf4ccdc7629aefa942a8f59d4d9aa615)
- Addressed review [r3908606737](https://github.com/hdmf-dev/hdmf-zarr/pull/325#discussion_r3908606737). `ZarrV2IO._decode_v2_dataset` fixed sparse v2 raw-array decoding. Missing chunks now retain the `fill_value` declared in the Zarr metadata instead of hard-coded defaults or uninitialized  np.empty  contents. See [c82b832](https://github.com/hdmf-dev/hdmf-zarr/pull/366/commits/c82b832221202da6c2095479e80d53b173876071)
- Addressed [r3908606675](https://github.com/hdmf-dev/hdmf-zarr/pull/325#discussion_r3908606675). Hardened Zarr v2 compatibility reads by rejecting unsafe pickle codecs by default to avoid accidental execution of code from untrusted files. Added explicit `allow_pickle=True` opt-in for `ZarrV2IO` for trusted legacy files across reads, namespace loading, and v2-to-v3 conversion. See [13c568d](https://github.com/hdmf-dev/hdmf-zarr/pull/366/commits/13c568d3173e7fd80c3113922485eba946ebf7ff)
- Updated minimum PyNWB version from `pynwb>=2.8.3` to `pynwb>=4.1.0` due to failing minimum dependency tests

## Checklist

- [ ] Does the PR clearly describe the problem and the solution?
- [ ] Does the PR use "Fix #XXX" notation to tell GitHub to close the relevant issue numbered XXX when the PR is merged?


rly:
> Ultimately, we may want to audit PyNWB/HDMF to remove reliance on these concepts of an array

This should be removed soon when https://github.com/NeurodataWithoutBorders/pynwb/pull/2235 is released. Let me know if you catch any others

oruebel:
> This should be removed soon when [NeurodataWithoutBorders/pynwb#2235](https://github.com/NeurodataWithoutBorders/pynwb/pull/2235) is released. Let me know if you catch any others

Thanks for the pointer. I'll give it try and let you know what I find.

oruebel:
> This should be removed soon when [NeurodataWithoutBorders/pynwb#2235](https://github.com/NeurodataWithoutBorders/pynwb/pull/2235) is released. Let me know if you catch any others

I tested locally, and it seems with updates in the V3 backend from #333 and the changes in PyNWB, all hdfm-zarr tests are passing. Also the HDMF Dev Branch Compatability CI tests are passing without the monkey patches. I'll test some more with the CI to see if there are any remaining issues or whether it will be sufficient to use the latest versions.

src/hdmf_zarr/backend.py
```diff
@@ -886,7 +863,9 @@ def write_attributes(self, **kwargs):
                         val = (
                             value.item()
                             if (isinstance(value, np.generic) and not isinstance(value, np.bytes_))
```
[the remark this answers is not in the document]
oruebel: Fixed in [488570d](https://github.com/hdmf-dev/hdmf-zarr/pull/366/commits/488570df01b6ccf88ff906410a3ad4a179b1a9b2)

h-mayorquin:
I ran the neuroconv test suite against this branch and the good news is that it mostly works. Some general comments:

The first is on reading files this branch writes. A compound column holding a reference comes back as a list of lists rather than a structured array, so `column.data[:]["idx_start"]` raises where it works on 0.13.0. Every icephys file is affected. It is not from your changes, the code is identical on `zarr-v3-migration`, so I opened #367 against base.

The second is on reading files written by the released stack. The v2 fallback decoder cannot handle a compound that holds an object field, which is how every reference column was written before v3, so those files do not open at all. I have a fix for it and will file it separately.

Separately, one big suggestion I have is to align the `ZarrDataIO` signature with the one on zarr. Concretely, we should change:

* `compressor` -> `compressors`, keeping `compressor=` as a deprecated alias
* `filters` keeps its name but now means `ArrayArrayCodec` only, which is not what it meant in v2
* add `serializer`, the `ArrayBytesCodec` slot, which today cannot be reached at all

This is a breaking release anyway, and aligning with the zarr API avoids forcing users to learn another API to configure their zarr datasets. Additionally, it is not as if `ZarrDataIO` is aligned with the HDF5 one, so there is no internal consistency cost either. Finally, the current API leads to the following confusions:

* `compressor` is singular but its type is `(BytesBytesCodec, list, bool)`, so it already takes a list under a singular name
* `Shuffle` is a `BytesBytesCodec` in v3 where it was a filter in v2, so `filters=[Shuffle(...)]` used to be right and is now wrong, and nothing in the signature says so
* the serializer is unreachable, so there is no way to say how strings are encoded

I hit this from the other side in neuroconv, where the configuration models carry an ordered `compressors` list and `get_data_io_kwargs()` has to flatten it back into a single compressor because that is all `ZarrDataIO` takes: https://github.com/catalystneuro/neuroconv/pull/1979

h-mayorquin:
> This should be removed soon when NeurodataWithoutBorders/pynwb#2235 is released. Let me know if you catch any others

@rly @oruebel already tested this and I did as well, so I will argue that we should drop most of it. At least we should drop `__len__`, `__iter__` and the scalar unwrap now rather than waiting for the release, so we don't build on promises that upstream has already said it will not hold (zarr-developers/zarr-python#3740, #3741 and #2529).

What remains trickier is the string handling, because of the release cycle, but I opened hdmf-dev/hdmf#1576 which should clear it out and then we can get rid of this entirely.

oruebel:
@rly @h-mayorquin A few more details on the investigation regarding `HDMFZarrArray`

> At least we should drop `__len__`, `__iter__`

There is a remaining issue in the gallery tests https://github.com/hdmf-dev/hdmf/issues/1579 that is caused by the ObjecMapper relying on `__iter__`. I've submitted a proposed fix to HDMF in https://github.com/hdmf-dev/hdmf/pull/1580. Once that is merged I believe we can drop  `__len__`, `__iter__`  in `HDMFZarrArray` here.

> What remains trickier is the string handling, because of the release cycle, but I opened [hdmf-dev/hdmf#1576](https://github.com/hdmf-dev/hdmf/pull/1576) which should clear it out and then we can get rid of this entirely.

Thanks @h-mayorquin for that PR. At least locally, tests are passing for me with that PR. I added a few additonal suggestions on that PR.

I think once https://github.com/hdmf-dev/hdmf/issues/1579  and https://github.com/hdmf-dev/hdmf/pull/1576 are released we may be able to drop  `HDMFZarrArray` here. The last issue I am still investigating is whether the unpacking of scalars is still need in `HDMFZarrArray.__getitem__()`. That is still causing additional test failures, but I have not looked closer yet, whether those are issues in HDMF or HDMF Zarr. For detail, here are the tests that are failing when removing the unwrapping of scalars via `HDMFZarrArray.__getitem__()`.

```
-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================== short test summary info ============================
FAILED tests/unit/test_io_convert.py::TestHDF5ToZarrDynamicTableC0::test_export_roundtrip
FAILED tests/unit/test_io_convert.py::TestZarrToHDF5DynamicTableC0::test_export_roundtrip
FAILED tests/unit/test_io_convert.py::TestZarrToZarrDynamicTableC0::test_export_roundtrip
FAILED tests/unit/test_io_convert.py::TestHDF5ToZarrDynamicTableC1::test_export_roundtrip
FAILED tests/unit/test_io_convert.py::TestZarrToHDF5DynamicTableC1::test_export_roundtrip
FAILED tests/unit/test_io_convert.py::TestZarrToZarrDynamicTableC1::test_export_roundtrip
FAILED tests/unit/test_io_convert.py::TestZarrToZarrNWBFile::test_export_roundtrip
FAILED tests/unit/test_io_convert.py::TestHDF5ToZarrNWBFile::test_export_roundtrip
FAILED tests/unit/test_io_convert.py::TestZarrToHDF5NWBFile::test_export_roundtrip
FAILED tests/unit/test_nwbzarrio.py::TestNWBZarrIO::test_read_nwb - TypeError...
FAILED tests/unit/test_nwbzarrio.py::TestNWBZarrIOCompoundDtype::test_compound_dtype_export_no_duplication
FAILED tests/unit/test_nwbzarrio.py::TestNWBZarrIOCompoundDtype::test_write_dataset_compound_dtype
FAILED tests/unit/test_zarrio.py::TestStringDatasetRead::test_1d_is_lazily_decoded_to_object_str
FAILED tests/unit/test_zarrio.py::TestHDMFZarrArray::test_zero_dimensional_result_is_unwrapped
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_electrodes_columns
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_electrodes_group_column
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_ephys_data_values
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_ephys_data_values_match_v2
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_ephys_exists
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_ephys_timestamps
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_experiment_description
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_exported_file_is_zarr_v3
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_identifier
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_instance_export_to_v3
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_institution
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_lab
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_n_devices
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_n_electrode_groups
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_n_electrodes
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_n_units
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_read_nwb_from_to_v3
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_session_description
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_session_id
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_session_start_time
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_subject_date_of_birth
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_subject_id
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_subject_species
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_timestamps_reference_time
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_units_exists
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_units_quality_column
ERROR tests/unit/test_zarrv2_backward_compat.py::TestV2ExportToV3::test_units_spike_times
===== 14 failed, 506 passed, 301 warnings, 27 errors in 115.35s (0:01:55) ======
```

h-mayorquin:
> The last issue I am still investigating is whether the unpacking of scalars is still need in `HDMFZarrArray.__getitem__()`

I took that one. It is still needed today. 12 of the 14 failures and all 27 errors come from three call sites in hdmf and pynwb that assume `array[0]` is a scalar, one line each, none of them here. The other 2 are the tests you added with the wrapper, `test_zero_dimensional_result_is_unwrapped` and `test_1d_is_lazily_decoded_to_object_str`. Those assert the unwrap, so they go with it.

Putting that together with what you already have, this is everything left before `HDMFZarrArray` can go. @rly's PR is out in 4.2.0 so it is not on the list:

* `__len__` and `__iter__`: @oruebel's PR, https://github.com/hdmf-dev/hdmf/pull/1580
* the scalar unwrap in `__getitem__`. These three are all of it, I checked that it comes out clean with them applied:
  * `EnumData` uses the elements as dict keys: https://github.com/hdmf-dev/hdmf/pull/1581
  * `file_create_date` is parsed one element at a time: https://github.com/NeurodataWithoutBorders/pynwb/pull/2263
  * `__apply_string_type` iterates a rank-0 array, which is where your 27 errors come from. I reproduced them and they survive #1580 on its own, so I left a suggestion on the line it changes: https://github.com/hdmf-dev/hdmf/pull/1580#discussion_r3964606893
* the string handling, so `dtype`, the `astype(object)` in `__getitem__` and `__array__`: https://github.com/hdmf-dev/hdmf/pull/1576, and then https://github.com/hdmf-dev/hdmf/issues/1578 for the read path

And a release of hdmf and pynwb with those.

Now that we have a full road map for this I think it is fine to release the zarr version 3 with the wrapping layer though. We have a plan to remove it bit by bit until we don't need it. The one thing I do think is worth settling before the release is aligning with the new zarr API as that will be breaking if we release and want to fix later, and I opened that separately:
https://github.com/hdmf-dev/hdmf-zarr/pull/369

oruebel:
> I took that one. It is still needed today.
> Putting that together with what you already have, this is everything left before `HDMFZarrArray` can go

Thank you @h-mayorquin for investigating this further and for the summary of the roadmap 👍 

> Now that we have a full road map for this I think it is fine to release the zarr version 3 with the wrapping layer though

I agree. Now that we have a roadmap, I think it is fine to move this forward with the full `HDMFZarrArray` in place as a temporary solution while HDMF and PyNWB are being updated. We can then remove `HDMFZarrArray` once the required PRs on HDMF/PyNWB have been released. 

I updated the docs of `HDMFZarrArray` in [9c90f80](https://github.com/hdmf-dev/hdmf-zarr/pull/366/commits/9c90f8041f90034c6d6d1411e6092e6174631217) to document the roadmap in the code and to clarify that downstream libraries should not rely on `HDMFZarrArray`.

src/hdmf_zarr/utils.py
```diff
@@ -549,10 +640,12 @@ def from_h5py_dataset(h5dataset, **kwargs):
         if isinstance(fillval, bytes):  # bytes are not JSON serializable so use string instead
             fillval = fillval.decode("utf-8")
         chunks = h5dataset.chunks if "chunks" not in kwargs else kwargs.pop("chunks")
+        compressors = kwargs.pop("compressor", compressors if compressors else None)
```
h-mayorquin: After my PR "Align `ZarrDataIO` with the new zarr version 3 codec API" (#369) `ZarrDataIO` does not take `compressor` anymore, the argument is `compressors`. So I think this is wrong. Passing `compressors=` here raises `got multiple values for keyword argument 'compressors'` as the value goes in twice, once from this line and once through `kwargs`, and the only name that gets through is the old one, the one the docstring above does not mention. The test passes because it uses `compressor=False`.

We should:
```suggestion
        compressors = kwargs.pop("compressors", compressors if compressors else None)
```

tests/unit/test_zarrdataio.py
```diff
@@ -262,6 +262,25 @@ def test_from_h5py_dataset_bytes_fillvalue(self):
...
+        h5file = h5py.File(self.hdf_filename, mode="a")
+        h5dset = h5file.create_dataset(
+            name="test",
+            data=np.arange(1000).reshape((10, 100)),
+            compression="gzip",
+            compression_opts=6,
+            shuffle=True,
+        )
+        filters = [TransposeCodec(order=(1, 0))]
+        re_zarrdataio = ZarrDataIO.from_h5py_dataset(h5dset, compressor=False, filters=filters)
```
h-mayorquin: Same as above, this is the old name.

```suggestion
        re_zarrdataio = ZarrDataIO.from_h5py_dataset(h5dset, compressors=False, filters=filters)
```

h-mayorquin (commented):
This looks good to me, some inline comments:

I also opened a PR against Ben's branch for the silent truncation on compound string fields (a longer value written into an existing compound was cut to the frozen width without an error), this branch gets it on the next merge:
https://github.com/hdmf-dev/hdmf-zarr/pull/370

One more that I can't attach inline as `pyproject.toml` is not in the diff:

https://github.com/hdmf-dev/hdmf-zarr/blob/ea5c8a745f961e48915bbde9743300c9627e3e0c/pyproject.toml#L35

`np.dtypes.StringDType` was added in numpy 2.0 but the pin here is `numpy>=1.26.0`. Before this PR only the write path used it. Now `HDMFZarrArray.dtype` calls it on every read, so a user with numpy 1.26 fails on the first array they touch. I think we should bump the floor.

```toml
    "numpy>=2.0",
```

src/hdmf_zarr/backend_zarrv2.py
```diff
@@ -244,7 +318,7 @@ def _make_spec_reader(cls, ns_group):
         return ZarrV2SpecReader(ns_group)
```
h-mayorquin: unused now that `_load_namespaces` builds `ZarrV2SpecReader` directly

oruebel:
> I think we should bump the floor.

Fixed in [b38b7e3](https://github.com/hdmf-dev/hdmf-zarr/pull/366/commits/b38b7e35dd81cbfaa756f27b55df369c7349786d)

h-mayorquin (approved):
LGTM

src/hdmf_zarr/backend_zarrv2.py
oruebel: Fixed in [34f9b09](https://github.com/hdmf-dev/hdmf-zarr/pull/366/commits/34f9b0977b44dc5353e4ef969df168f4862f716a) .  Removed `_make_spec_reader` from both `ZarrIO` and `ZarrV2IO` since the overridden `_load_namespaces` takes care of constructing the spec reader directly.

PR #367, opened 2026-09-02
Rebuild a resolved compound row as a record instead of a list

h-mayorquin:
This has #325 as a base.

This comes from me running @oruebel's branch (#366) on the neuroconv testing suite locally. There it fails for `IntracellularRecordingsTable`. The reason is that the cell of both response and stimulus, which is the entry of a compound data type with shape `[('idx_start', '<i4'), ('count', '<i4'), ('timeseries', '<U512')]`, is never converted into such and instead kept as a list.

This makes this fail with `column.data[:]["idx_start"]`:

```
TypeError: list indices must be integers or slices, not str
```

This used to work in place. On `dev` the reference field is `|O` on disk, so `__swap_refs` could assign the resolved `TimeSeries` straight into the record. In v3 that field is `<U512`, a string field which does not work directly so I think it defaulted to wrapping the row in a list to make the assignment legal. What is missing is the conversion back from string to the object. This PR builds a resolved dtype, the same as the on-disk one but with `object` in the reference fields, once at read time, so the rows are already the right size and the resulting array is compound dtype already. This is consistent with what the HDF5 backend returns.

so we get:
```
base branch   d[0]: list   is np.void: False   d[:]: list      dtype.names: None
with this PR  d[0]: void   is np.void: True    d[:]: ndarray   dtype.names: ('idx_start', 'count', 'timeseries')
HDF5 twin     d[0]: void   is np.void: True    d[:]: ndarray   dtype.names: ('idx_start', 'count', 'timeseries')
```

The test passes unmodified on `dev`, so this is a regression rather than a new guarantee: `dev` returns a structured array here and the migration silently started returning lists.

This does not require @oruebel's PR and is a simple fix so I opened with #325 as base. I added a test so we don't fail on this again.

### DIFF
--- a/src/hdmf_zarr/zarr_utils.py
+++ b/src/hdmf_zarr/zarr_utils.py
@@ -180,6 +180,16 @@ class AbstractZarrTableDataset(DatasetOfReferences):
                 tmp.append(sub.type.__name__)
         self.__dtype = tmp
 
+        # The dtype a resolved row has in memory. Zarr v3 stores a reference inside a compound as a
+        # fixed-width string, which cannot hold the resolved Container or Builder, so the reference
+        # fields are widened to object here. This is also the dtype the HDF5 backend hands back.
+        self.__resolved_dtype = np.dtype(
+            [
+                (name, object if types[index] == DatasetBuilder.OBJECT_REF_TYPE else self.dataset.dtype[index])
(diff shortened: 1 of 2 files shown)


oruebel (approved):
Looks good to me

h-mayorquin:
Thanks for the review.

PR #368, opened 2026-09-09
Decode a zarr v2 compound that holds an object field

h-mayorquin:
I found this running this branch against the neuroconv test suite. The icephys files I wrote with the released stack, hdmf-zarr 0.13.0 on zarr 2, cannot be opened at all. The `stimulus` and `response` columns of `IntracellularRecordingsTable` are compound with an object field, `[['idx_start', '<i4'], ['count', '<i4'], ['timeseries', '|O']]`, so zarr v3 refuses them and `_decode_v2_dataset` takes over. There the dtype arrives from JSON as a list of lists where `np.dtype` wants tuples, and `is_object` matches only an exact `|O` instead of a dtype that contains an object field. `_iter_children` downgrades both to a warning and drops the column, so the failure surfaces much later as a `ConstructError` about `colnames`. This PR builds the dtype from tuples, treats "has an object field" as object, and resolves the structured `fill_value` that v2 stores as base64 of a pickled scalar to a zeroed record, the way zarr v2 itself does.

I am sending this without a test because I would like to know how you want it covered:

* Add an intracellular recording to the v2 fixture in `tests/unit/helpers`, as it has no compound dataset in it at all. That covers this end to end, but regenerating rewrites every `object_id` and the cached schema with it, so the diff is 171 files no matter what I pin in the generation venv.
* Build a synthetic v2 store in the test the way `TestV2ObjectChunkDecoding` already does. That stays small but only covers the decoder.

To be honest, my preference is the second one. The v2 path is only there for files people already have, so the decoder is what I want guarded, and I don't think there are enough v2 files around to justify 171 files of fixture churn every time we cover one more shape. But this is a one-off thing (only one v2 to v3 migration) and I am fine with whatever you guys want to move forward with.

### DIFF
--- a/src/hdmf_zarr/backend_zarrv2.py
+++ b/src/hdmf_zarr/backend_zarrv2.py
@@ -500,6 +500,11 @@ class ZarrV2IO(ZarrIO):
         is_object = dtype == np.dtype("|O") or has_object_field
         result_dtype = dtype if has_object_field else (object if is_object else dtype)
         fill_value = zarray_meta.get("fill_value", None if is_object else 0)
+        if has_object_field and isinstance(fill_value, str):
+            # A structured fill for an object-codec array is stored as base64 of a pickled scalar.
+            # Reading it back would need the pickle codec this backend gates, and zarr v2 resolves
+            # the one hdmf-zarr writes to 0, which for a record means every field zeroed.
+            fill_value = 0
         ndim = len(shape)
         chunk_grid = tuple((s + c - 1) // c for s, c in zip(shape, chunks))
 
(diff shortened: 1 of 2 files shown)


oruebel:
> To be honest, my preference is the second one.

@h-mayorquin that sounds good to me. Please add the test and then we can merge this PR with parent.

h-mayorquin:
> > To be honest, my preference is the second one.
> 
> @h-mayorquin that sounds good to me. Please add the test and then we can merge this PR with parent.

Should I update the fixture or create a unit test that covers this? see my question in the first message.

oruebel:
> Should I update the fixture or create a unit test that covers this? see my question in the first message.

If I understand your comment correctly, then your preferences is to _"> * Build a synthetic v2 store in the test the way `TestV2ObjectChunkDecoding` already does. That stays small but only covers the decoder."_ in order to keep the changes focused and small. That sounds reasonable to me.

h-mayorquin:
@oruebel I have added a test now.

oruebel (approved):

PR #373, opened 2026-09-10
Fix #371 Increase Zarr and Python minimum versions

### LINKED ISSUE #371: Zarr v3 does not work with python 3.11
From zarr v3.1 to zarr v3.2 the way that structured dtypes are handled changed:

* zarr 3.1.x writes a compound dataset with the data type name `structured`, fields as `[name, dtype]` pairs, and warns that this representation has no specification.
* zarr 3.2.0 and later write `struct`, the name that zarr-extensions registered in March, and 3.3.0 dropped the warning:
https://github.com/zarr-developers/zarr-extensions/pull/45

The new name is not backwards compatible. This is zarr 3.1.5 opening a compound written by 3.3.0:

```
ValueError: No Zarr data type found that matches {'name': 'struct', 'configuration': {'fields': [...]}}
```

The other direction only warns about a missing `endian` key.

The problem is that "Migrate hdmf-zarr from zarr-python v2 to v3" (#325) pins `zarr>=3.1.3` which uses the old way. The obvious solution would be to put the pin higher but this has a problem, zarr 3.2.0 and later only support python 3.12. Python 3.11 is in security support until 2027-10:
https://devguide.python.org/versions/

Now another solution is to have different pins per python version:

```toml
"zarr>=3.3.0; python_version >= '3.12'",
"zarr>=3.1.3; python_version < '3.12'",
```

I checked this and it does not work because python 3.11 will not be able to read files produced by python 3.12 which I think is the wrong error to pass to the users. Any NWB file with an `IntracellularRecordingsTable` or a `TimeIntervals` with time series references written on 3.12 fails to open on 3.11, with nothing at write time saying so.

Honestly to me it seems that we should only support hdmf-zarr with python 3.12, what is the downside? zarr kind of already made this decision for us.

@bendichter @oruebel @rly what do you think? Is there something I am missing?

oruebel:
Fix #371 

1. __`pyproject.toml`__:

   - Updated `requires-python` from `>=3.11` to `>=3.12`.
   - Removed the Python 3.11 Trove classifier.
   - Updated the Zarr version pin from `zarr>=3.1.3` to `zarr>=3.3.0`.

2. __`tox.ini`__:

   - Dropped `py311` from the main `envlist` and environment groupings (e.g. replacing `py{310,311,312,313,314}` with `py{312,313,314}`).
   - Renamed and reconfigured the `py311-minimum` and `gallery-py311-minimum` test environments to run on Python 3.12 (`py312-minimum` and `gallery-py312-minimum`).

3. __GitHub Actions (`.github/workflows/run_tests.yml` & `.github/workflows/run_all_tests.yml`)__:

   - Purged all `python3.11` test entries from the matrices across all platforms (Ubuntu, Windows, and macOS).
   - Changed the `-minimum` dependencies jobs (e.g. `linux-python3.11-minimum`, `conda-linux-python3.11-minimum`, etc.) to run on Python 3.12 (`*python3.12-minimum`).

4. __Bug Report Issue Template (`.github/ISSUE_TEMPLATE/bug_report.yml`)__:
   - Removed `"3.11"` and `"3.10"` from the available options in the Python version dropdown.

## Checklist

- [X] Did you update `CHANGELOG.md` with your changes?
- [X] Does the PR clearly describe the problem and the solution?
- [X] Have you reviewed our [Contributing Guide](https://github.com/hdmf-dev/hdmf-zarr/blob/dev/docs/CONTRIBUTING.rst)?
- [X] Does the PR use "Fix #XXX" notation to tell GitHub to close the relevant issue numbered XXX when the PR is merged?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -56,8 +56,6 @@ body:
         - "3.14"
         - "3.13"
         - "3.12"
-        - "3.11"
-        - "3.10"
         - older (please specify in "Package Versions" below)
     validations:
       required: true
--- a/.github/workflows/run_all_tests.yml
+++ b/.github/workflows/run_all_tests.yml
@@ -25,22 +25,19 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - { name: linux-python3.11-minimum     , test-tox-env: py311-minimum   , python-ver: "3.11", os: ubuntu-latest }
-          - { name: linux-python3.11             , test-tox-env: py311           , python-ver: "3.11", os: ubuntu-latest }
+          - { name: linux-python3.12-minimum     , test-tox-env: py312-minimum   , python-ver: "3.12", os: ubuntu-latest }
           - { name: linux-python3.12             , test-tox-env: py312           , python-ver: "3.12", os: ubuntu-latest }
           - { name: linux-python3.13             , test-tox-env: py313           , python-ver: "3.13", os: ubuntu-latest }
           - { name: linux-python3.14             , test-tox-env: py314           , python-ver: "3.14", os: ubuntu-latest }
           - { name: linux-python3.14-optional    , test-tox-env: py314-optional  , python-ver: "3.14", os: ubuntu-latest }
           - { name: linux-python3.14-prerelease  , test-tox-env: py314-prerelease, python-ver: "3.14", os: ubuntu-latest }
-          - { name: windows-python3.11-minimum   , test-tox-env: py311-minimum   , python-ver: "3.11", os: windows-latest }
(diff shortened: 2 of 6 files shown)


bendichter (approved):

h-mayorquin (commented):
LGTM

One thing, the minimum we require is `3.2.0`, where `structured` became `struct`. I think `3.3.0`, which
you use here, might be better anyway because of the `consolidated_metadata` fix for empty groups.

One thing though, dandi-cli cannot upload NWB Zarr and caps zarr at `<=3.1.5`. My only concern is that they
might have a reason for that and we might not be able to meet them there, but I guess this is fine and we
can update if and when they answer: https://github.com/dandi/dandi-cli/issues/1920

rly:
Thanks for opening that issue on dandi-cli @h-mayorquin . I wouldn't have thought to check. It would be good to know from the dandi team the reason for their cap!

PR #374, opened 2026-09-10
Prevent data loss in parallel writes to sharded arrays

h-mayorquin:
[Edited after discussion]

When reviewing #372 I reproduced data loss from two workers updating different chunks in the same shard. After that, I found the same problem documented in [Zarr #3514](https://github.com/zarr-developers/zarr-python/issues/3514). [Dask addressed it by aligning its write tasks to shards](https://github.com/dask/dask/pull/12105). Before this change the parallel queue created one task per iterator buffer. Those buffers can overlap the same shard even when their array selections don't overlap. A partial update reads the existing encoded shard and writes its replacement, so one worker can overwrite the other worker's changes.

This PR prevents different workers from updating the same shard. A worker can be assigned to work with several complete shards, so an aligned buffer is still read and written in one assignment. Workers still operate in parallel but across different shards:

image

When buffers don't align with shards, one task owns each shard and processes its buffer pieces sequentially.

An important note: the implementation here is flexible and does not couple buffer shapes to shard shapes as those come from different concerns (storage layout in the destination vs reading patterns from the source). In practice though, **we should recommend buffers equal to the shard shape or an integer multiple along each axis**. The critical point is that a partial shard update reads the whole existing encoded shard into memory and assembles its replacement. As confirmed in [this maintainer reply](https://github.com/zarr-developers/zarr-python/issues/3604#issuecomment-3608753457), Zarr-Python has no API to defer writing a shard until all its chunks have been processed. I verified this behavior locally.

This means that we have two problems when the shard is larger than the buffer: 
1. The existing encoded shard and its replacement add memory beyond the source buffer. So `buffer_shape` doesn't bound total write memory, and a large shard can defeat the purpose of choosing a small buffer.
2. We repeatedly read and rewrite the data already stored in that shard as successive buffer pieces are written.

I have added documentation on this as I am unsure that python-zarr will change anytime soon and we can change it then.

I also added `shards="auto"` to `ZarrDataIO` and we pass it through.

### DIFF
--- a/src/hdmf_zarr/utils.py
+++ b/src/hdmf_zarr/utils.py
@@ -7,6 +7,7 @@ import math
 import json
 import logging
 import os
+from itertools import product
 from collections import deque
 from collections.abc import Iterable
 from typing import Optional, Union, Literal, Tuple, Dict, Any
@@ -229,25 +230,6 @@ class ZarrIODataChunkIteratorQueue(deque):
                     )
                     continue
 
-                # GenericDataChunkIterator tiles from zero in buffer_shape steps. Each
-                # task must own whole shards, except at the outer array boundary.
-                # Inspect the created array because Zarr may have chosen shards="auto".
-                shards = zarr_dataset.shards
-                if shards is not None and any(
-                    buffer_size % shard_size != 0 and buffer_size != array_size
-                    for buffer_size, shard_size, array_size in zip(iterator.buffer_shape, shards, zarr_dataset.shape)
-                ):
-                    warn(
-                        f"Writing dataset '{zarr_dataset.path}' sequentially: iterator buffer shape "
-                        f"{iterator.buffer_shape} is not aligned with shard shape {shards}. "
-                        "Parallel writes could update the same shard and lose data. "
-                        "Use a buffer shape that is a multiple of the shard shape, "
-                        "or covers the entire array dimension.",
-                        UserWarning,
-                        stacklevel=2,
-                    )
-                    continue
-
                 # Add this entry to a running list to remove after initial pass (cannot mutate during iteration)
                 parallelizable_iterators.append((zarr_dataset, iterator))
 
@@ -262,7 +244,15 @@ class ZarrIODataChunkIteratorQueue(deque):
                 progress_bar_options.update(**per_iterator_progress_options)
 
                 iterator_itemsize = iterator.dtype.itemsize
-                for buffer_selection in iterator.buffer_selection_generator:
+                # Preserve whole buffers when they own complete shards. Otherwise,
+                # give each shard one task, using Zarr's resolved layout for auto.
+                selections = (
+                    self._iter_shard_selections(zarr_dataset.shape, zarr_dataset.shards)
+                    if zarr_dataset.shards is not None
+                    and not self._buffers_contain_shards(zarr_dataset.shape, zarr_dataset.shards, iterator.buffer_shape)
+                    else iterator.buffer_selection_generator
+                )
+                for buffer_selection in selections:
                     store = zarr_dataset.store
                     store_path = str(store.root) if hasattr(store, "root") else str(store)
                     buffer_map_args = (store_path, zarr_dataset.path, iterator, buffer_selection)
@@ -392,6 +382,33 @@ class ZarrIODataChunkIteratorQueue(deque):
         _operation_to_run = operation_to_run
 
     @staticmethod
+    def _buffers_contain_shards(shape, shards, buffers):
+        """Return whether buffer boundaries avoid splitting destination shards."""
+        return all(buffer == size or buffer % shard == 0 for size, shard, buffer in zip(shape, shards, buffers))
+
+    @staticmethod
+    def _iter_shard_selections(shape, shards):
+        """Yield one non-overlapping task selection per destination shard."""
+        for start in product(*(range(0, size, shard) for size, shard in zip(shape, shards))):
+            yield tuple(slice(pos, min(pos + shard, size)) for pos, shard, size in zip(start, shards, shape))
+
+    @staticmethod
+    def _iter_shard_buffer_selections(shard_selection, buffer_shape):
+        """Intersect a shard with the iterator buffer grid without loading a full shard.
+
+        A buffer crossing shard boundaries is read in separate pieces by the shard
+        owners. Each requested piece is no larger than the iterator buffer shape.
+        """
+        starts = (
+            range(part.start // size * size, part.stop, size) for part, size in zip(shard_selection, buffer_shape)
+        )
+        for start in product(*starts):
+            yield tuple(
+                slice(max(pos, part.start), min(pos + size, part.stop))
+                for pos, size, part in zip(start, buffer_shape, shard_selection)
+            )
+
+    @staticmethod
     def _write_buffer_zarr(
         worker_context: Dict[str, Any],
         zarr_store_path: str,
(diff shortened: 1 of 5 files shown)


src/hdmf_zarr/utils.py
```diff
@@ -559,12 +577,13 @@ class ZarrDataIO(DataIO):
         },
         {
             "name": "shards",
-            "type": (list, tuple),
+            "type": (list, tuple, str),
```
oruebel, line 606: Instead of allowing `str` with only a single valid value of `auto`, wouldn't it be simpler to just allow a bool so that when `shards=true` it means auto. This would also be similar to how `chunks=true` enables auto chunking in HDF5.

oruebel:
@alejoe91 could you take a look at this

h-mayorquin:
Let me do some changes, I have think on another approach which I think is better.

h-mayorquin:
Ok, this was a rabbit hole. 

I pushed an approach that keeps parallel iterator writes when buffers don't align with shards. The previous version of this PR handled that case by reading and writing the dataset one iterator buffer at a time in a single HDMF process. That sequential fallback prevents the race, but it also gives up parallel iterator reads and writes across independent shards. To avoid the race we only need to prevent different tasks from writing to the same shard.

So each shard belongs to one task. When buffers contain complete shards, we keep the existing buffer tasks. Otherwise, we assign one task per shard and read the portions of iterator buffers inside it. Writes within a shard are sequential and different shards can be written in parallel.

The first diagram shows the aligned case. Buffers and shards both have shape `(8000, 384)`, with four `(2000, 384)` chunks per shard. Each buffer is one task with one source read and one Zarr assignment. A buffer containing several complete shards would also stay one task:

image

The second diagram uses `(6000, 384)` buffers and `(8000, 384)` shards. Task A reads and writes `0:6000`, then `6000:8000`, while another worker can handle the next shard. Each read requests only the intersection and its assignment finishes before the next read:

image


Each assignment is `array[selection] = piece`. Zarr handles chunk encoding and shard storage. The queue uses the created array's `shards`, including with `shards="auto"`. Each source read stays within `buffer_shape`, although Zarr can use additional memory internally. I documented buffers equal to shards or integer multiples along each axis as the preferred pattern. Other shapes remain supported, but partial updates can repeatedly read and rewrite existing shard contents. That cost was already present before this change.

I compared this with the sequential fallback. The performance benefit depends on the buffer and shard layout. The fallback still benefits from Zarr's internal concurrency.

I updated the tests for the new scheduling. Let me know what do you think.

src/hdmf_zarr/utils.py
h-mayorquin: I am following the philosophy of respecting the backend API rather than trying to invent our own. This avoids users having to re-learn new terminology. HDF5 and Zarr already differ in how they expose compression: Zarr uses `compressors`, while h5py uses `compression` and `compression_opts`. How do you feel about that?

Personally, I don't like mixing types in the same argument here but I also don't like bools that end up needing more flags and combinations when an enum could represent the modes (see [boolean explosion](https://stately.ai/blog/2020-05-27-state-management-bad-bolean-good-boolean#implicit-states)) so I am torn. Rather than designing this myself I would rather just stick to what Zarr already does. I think is the path of least friction.

tests/unit/test_parallel_write.py
```diff
@@ -279,3 +284,132 @@ def test_extra_keyword_argument_propagation(tmpdir):
...
+            items = list(iterable)
+            tasks.extend(items)
+            return super().map(fn, items, **kwargs)
+
+    data = np.arange(1, np.prod(shape) + 1, dtype="int32").reshape(shape)
+    iterator = PickleableDataChunkIterator(data, chunk_shape=chunks, buffer_shape=buffers)
+    column = VectorData(name="values", description="", data=ZarrDataIO(iterator, chunks=chunks, shards=shards))
+    table = DynamicTable(name="table", description="", id=list(range(shape[0])), columns=[column])
+    store = str(tmp_path / "data.zarr")
+    with patch("hdmf_zarr.utils.ProcessPoolExecutor", wraps=RecordingExecutor) as executor:
```
alejoe91: Why do you need to patch this?

tests/unit/test_parallel_write.py
```diff
@@ -279,3 +284,132 @@ def test_extra_keyword_argument_propagation(tmpdir):
...
+
+
+@pytest.mark.parametrize(
+    "shape,chunks,shards,buffers",
+    [
+        ((8,), (2,), (8,), (4,)),
+        ((22,), (2,), (8,), (6,)),
+        ((22, 14), (2, 2), (8, 8), (6, 10)),
+        ((18,), (2,), (8,), (8,)),
+        ((34,), (2,), (8,), (16,)),
```
alejoe91: If I unserstand correctly, these two cases (plus the last with no shards) are the only ones that will support parallel write. Correct?

- `((18,), (2,), (8,), (8,))`: 1 shard per job
- `((34,), (2,), (8,), (16,))`: 2 shards per job

tests/unit/test_parallel_write.py
alejoe91: Ah I see you want to record the `tasks`

src/hdmf_zarr/utils.py
alejoe91: I agree with @h-mayorquin . Let's keep "auto"

alejoe91:
> Ok, this was a rabbit hole.
> 
> I pushed an approach that keeps parallel iterator writes when buffers don't align with shards. The previous version of this PR handled that case by reading and writing the dataset one iterator buffer at a time in a single HDMF process. That sequential fallback prevents the race, but it also gives up parallel iterator reads and writes across independent shards. To avoid the race we only need to prevent different tasks from writing to the same shard.
> 
> So each shard belongs to one task. When buffers contain complete shards, we keep the existing buffer tasks. Otherwise, we assign one task per shard and read the portions of iterator buffers inside it. Writes within a shard are sequential and different shards can be written in parallel.
> 
> The first diagram shows the aligned case. Buffers and shards both have shape `(8000, 384)`, with four `(2000, 384)` chunks per shard. Each buffer is one task with one source read and one Zarr assignment. A buffer containing several complete shards would also stay one task:
> 
> image
> The second diagram uses `(6000, 384)` buffers and `(8000, 384)` shards. Task A reads and writes `0:6000`, then `6000:8000`, while another worker can handle the next shard. Each read requests only the intersection and its assignment finishes before the next read:
> 
> image
> Each assignment is `array[selection] = piece`. Zarr handles chunk encoding and shard storage. The queue uses the created array's `shards`, including with `shards="auto"`. Each source read stays within `buffer_shape`, although Zarr can use additional memory internally. I documented buffers equal to shards or integer multiples along each axis as the preferred pattern. Other shapes remain supported, but partial updates can repeatedly read and rewrite existing shard contents. That cost was already present before this change.
> 
> I compared this with the sequential fallback. The performance benefit depends on the buffer and shard layout. The fallback still benefits from Zarr's internal concurrency.
> 
> I updated the tests for the new scheduling. Let me know what do you think.

Could you update the PR description with the new strategy?

src/hdmf_zarr/utils.py
```diff
@@ -244,7 +244,15 @@ def exhaust_queue(self):
                 progress_bar_options.update(**per_iterator_progress_options)
 
                 iterator_itemsize = iterator.dtype.itemsize
-                for buffer_selection in iterator.buffer_selection_generator:
+                # Preserve whole buffers when they own complete shards. Otherwise,
+                # give each shard one task, using Zarr's resolved layout for auto.
```
alejoe91, line 248:
```suggestion
                # Preserve whole buffers when they own complete shards. Otherwise,
                # give each shard one task, using Zarr's resolved layout for auto.
                # If a shard owns multiple buffers, they will be iterated over within the 
                # same task.
```
Or somethig similar :)

alejoe91 (approved):
Just one minor point to extend a comment to make the process easier

alejoe91:
And another question / suggestion: in case `shards="auto"`, should we set the `shard` shape externally to match the buffer shape to maximize write efficiency?

src/hdmf_zarr/utils.py
h-mayorquin: Done.

h-mayorquin:
@alejoe91 I updated the PR description and title. Regarding shards="auto", I would keep it as a pass-through so users get Zarr's native behavior. I don't think we should give it a different meaning in HDMF. Users who want shards to match their buffers can set that explicitly.

oruebel (approved):

PR #382, opened 2026-09-16
Fill CHANGELOG gaps and close doc and test gaps from the v3 migration review

rly:
Non-blocking cleanup from the review of #325. Every item here is documentation, changelog, or test coverage, plus one small path-handling fix. No behavior that #325 introduces is changed.

## Docs

`README.rst` is the PyPI long description and still described hdmf-zarr as a Zarr v2 backend, with a line saying Zarr v3 support is planned and pointing at #202 as open work. #325 removed exactly that text from `docs/source/index.rst` but `README.rst` was never touched, so the 0.14 release page would have contradicted the release. It now states which classes read and write v3 and which read legacy v2.

`docs/source/storage.rst:125` still named `DirectoryStore`, which does not exist in zarr-python 3. The same sentence in `overview.rst` was already updated to `LocalStore` in #325, so the two pages disagreed.

The `tox.ini` note explaining the interpreter-from-name convention was illustrated with `py311`, which `envlist` no longer contains.

## CHANGELOG

The 0.14 notes were missing several user-visible breaks that a downstream maintainer needs in order to plan an upgrade:

- `synchronizer` and `object_codec_class` are gone from the `ZarrIO`/`NWBZarrIO` constructors and properties. Passing either raises `TypeError: ZarrIO.__init__: unrecognized argument: '<name>'`, which I confirmed against the branch.
- `storage_options` is accepted only with `mode="r"`, so writing directly to a remote store raises `ValueError`.
- `DirectoryStore`, `NestedDirectoryStore`, and `TempStore` no longer exist, so code constructing one needs `LocalStore`.
- The dependency bullet named only Python and zarr, while the release also moves `hdmf` to `>=6.2.0`, `numpy` to `>=2.0.0`, `pynwb` to `>=4.2.0`, and drops the `numcodecs<0.16.0` cap. The two-major-version hdmf jump is the item most likely to block pynwb, neuroconv, and NWB GUIDE.
- Files written by 0.14 cannot be opened by 0.13 or earlier. The notes covered only the reverse direction.

## `ZarrSpecReader.source`

`ZarrSpecReader` built its source by applying `os.path.abspath` to the `file://` URL that `str(LocalStore)` returns. Because that string does not start with `/`, it was joined with the cwd, giving a source like `<cwd>/file:/abs/path/f.nwb.zarr:/specifications` that names no real file and embeds the working directory.

Nothing in hdmf opens or stats that value, so the effect is narrow: it is used as a dict key in hdmf's `__included_specs` bookkeeping, and the embedded cwd makes the "already included" check miss when the same file is read twice with a `chdir` in between, so `load_namespaces` returns `{}` instead of the dependency dict.

The `LocalStore`-aware logic that `ZarrIO._get_store_path` already had moves to `utils.get_store_path`, which `ZarrSpecReader` needs and which cannot live in `backend.py` without a circular import. `_get_store_path` is defined once and overridden nowhere, so rather than leave it as a pass-through the 16 call sites in `backend.py` and `backend_zarrv2.py` call the helper directly.

Its docstring claimed remote stores return "the string representation", which reads as a URL. `FsspecStore` defines no `__str__` (`zarr/storage/_fsspec.py:265`), so `str()` falls back to `__repr__` and yields `<FsspecStore(S3FileSystem, bucket/path)>`. `LocalStore` gives `file://...` and `MemoryStore` gives `memory://<object id>`. The docstring now names what each store class produces rather than implying a common format.

**Worth a look from the #325 authors, not changed here:** `backend.py` sets every `GroupBuilder.source` and `DatasetBuilder.source` from this function on each read, so reading a file from S3 gives every builder a source of `<FsspecStore(...)>`. `ZarrIO.__init__` avoids this for string paths by skipping `os.path.abspath` when `"://"` is present, but the read path always goes through the store object.

## Tests

`test_compression_info.py` ended every one of its three tests with `assertIsInstance(str(array.info), str)` and `assertTrue(len(info_str) > 0)`, both true by construction. The file was originally the regression test for `nbytes_stored` returning a placeholder under consolidated metadata, and its real assertions were dropped during the migration because zarr v3 moved stored size from `Array.info` to `Array.info_complete()`. The tests now assert that the Blosc codec configured on `ZarrDataIO` survives the write and read back, and that `nbytes_stored()` is above zero and below `nbytes`.

`NWBZarrIO.read_nwb` sets `storage_options=dict(anon=True)` for `s3://` paths. Its only coverage was `TestFSSpecStreaming.test_fsspec_streaming_via_read_nwb`, which #325 redirected to `NWBZarrV2IO`, a separate implementation. Two tests now pin the `storage_options` passed for an `s3://` path and for a local path, without needing network access.

`test_get_store_path_shallow` and `test_get_store_path_deep` asserted only `assertIsInstance(path, str)`, true for every possible return value, and carried a comment describing a `str(store)` return that the code does not produce. They now assert the resolved filesystem path.

Every assertion added here was verified by mutation rather than by inspection:

| Mutation | Result |
| --- | --- |
| `ZarrDataIO` drops the configured `compressors` | all 3 compression tests fail (`expected one BloscCodec, got ()`) |
| `nbytes_stored()` returns `-1` | both consolidated-metadata tests fail (`-1 not greater than 0`) |
| `read_nwb` loses the `anon=True` branch | the s3 test fails, the local test stays green |
| `get_store_path` returns `str(store)` | both store-path tests fail (`'file://test_io.zarr' != '/private/.../test_io.zarr'`) |

## Verification

`pytest tests/` goes from 579 to 581 passed, 14 skipped.


CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ This release marks a major transition for `hdmf-zarr`, updating the core storage
...
+
+### Removed
+- **`synchronizer` argument:** `ZarrIO` and `NWBZarrIO` no longer accept a `synchronizer` argument, and the `ZarrIO.synchronizer` property is removed. zarr-python v3 does not provide the synchronizer mechanism this wrapped. Passing `synchronizer=...` raises `TypeError`.
+- **`object_codec_class` argument:** `ZarrIO` and `NWBZarrIO` no longer accept an `object_codec_class` argument, and the `ZarrIO.object_codec_class` property is removed. References, cached specs, and compound datasets are serialized as JSON in `StringDType` arrays, so there is no object codec to select. Passing `object_codec_class=...` raises `TypeError`.
+- **Store classes:** `SUPPORTED_ZARR_STORES` covers `LocalStore`, `FsspecStore` when fsspec is installed, and any other zarr v3 `Store` subclass. `DirectoryStore`, `NestedDirectoryStore`, and `TempStore` do not exist in zarr-python v3, so code that constructs one needs to use `LocalStore` instead.
+- **Remote write:** Passing `storage_options` is supported only with `mode="r"`. Writing an NWB Zarr file directly to a remote store is not currently supported and raises `ValueError`.
```
oruebel: I would move this part to the "### Changes" section. In particular the last tow items (store classes an remote write) do not actually discuss removal of arguments but changes in behavior

CHANGELOG.md
oruebel, line 40: Add this PR to the list of pull requests related to this release

CHANGELOG.md
```diff
@@ -16,6 +23,7 @@ The core functionality has been overhauled to transition to Zarr V3, including s
 * **Aligned `ZarrDataIO` with the Zarr v3 codec API:** The `compressor` argument is renamed to `compressors`, the `serializer` (`ArrayBytesCodec`) slot is now directly accessible, and `filters` only applies to `ArrayArrayCodec`. @h-mayorquin [#369](https://github.com/hdmf-dev/hdmf-zarr/pull/369)
 * **Support dataset sharding via `ZarrDataIO`:** Added new `shards` argument to `ZarrDataIO` to define custom sharding properties, including support for automatic sharding and guarded parallel iterator writes. @h-mayorquin @alejoe91 [#374](https://github.com/hdmf-dev/hdmf-zarr/pull/374) [#372](https://github.com/hdmf-dev/hdmf-zarr/pull/372)
 * **Python and Zarr Requirements:** Bumped minimum required Python version to 3.12 and Zarr dependency to `>=3.3.0` due to backwards-incompatible changes in Zarr v3 structured data types. @oruebel [#373](https://github.com/hdmf-dev/hdmf-zarr/pull/373)
+* **Other Dependency Requirements:** Bumped `hdmf` to `>=6.2.0`, `numpy` to `>=2.0.0`, and `pynwb` to `>=4.2.0`, and removed the `numcodecs<0.16.0` upper bound.
```
oruebel: I think those two entries could be merged into a single item on "Requirements Changes:"

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ This release marks a major transition for `hdmf-zarr`, updating the core storage
 - **Zarr V3 Standard:** The primary `ZarrIO` and `NWBZarrIO` classes now write and read Zarr v3 format exclusively, using `zarr-python` v3.
 - **Legacy Zarr V2 Support:** Legacy Zarr v2 files cannot be read by `ZarrIO` directly. Instead, use the newly added `ZarrV2IO` and `NWBZarrV2IO` backends to read them. Attempting to open a v2 file with `ZarrIO` will now raise a clear error directing users to the v2 classes.
 - **Migration Path:** To migrate existing files to the new format, you can use `NWBZarrV2IO.export_to_v3(...)` or the one-shot `NWBZarrV2IO.convert_to_v3(...)` static helper to efficiently convert NWB Zarr v2 files to Zarr v3.
+- **Forward Compatibility:** Files written by this release are Zarr v3 and cannot be opened by hdmf-zarr 0.13 or earlier, which require `zarr<3`. Collaborators who need to read a file written with 0.14 must also upgrade to 0.14.
```
```suggestion
- **Forward Compatibility:** Files written by this release are Zarr v3 and cannot be opened by `hdmf-zarr` 0.13 or earlier, which require `zarr<3`. Upgrade to `hdmf-zarr` 0.14 in order to read Zarr v3 files written with `hdmf-zarr`.
```

CHANGELOG.md
rly: Done. The `### Removed` section is gone and all four items moved into `### Changes`, with the two behavior items renamed to `Supported store classes` and `Remote write` so they no longer read as argument removals.

CHANGELOG.md
rly: Done, added to the list.

CHANGELOG.md
rly: Done, merged into a single `Requirements Changes` item covering Python, Zarr, hdmf, numpy, pynwb and the dropped numcodecs cap.

CHANGELOG.md
rly: Applied as suggested.

oruebel (approved):

PR #388, opened 2026-09-16
Enforce the pickle policy on every codec a v2 array declares

### LINKED ISSUE #385: `allow_pickle=False` does not prevent pickle execution when a Zarr v2 file lists another object codec first
## Summary

Reading a Zarr v2 file can require decoding data that was stored with Python's `pickle`. Unpickling data from a file you did not write is equivalent to running code from that file, so the v2 reader gates it behind `allow_pickle`, which defaults to `False`.

The gate is applied in the wrong place. It guards only the project's own fallback decoder, which runs when zarr-python cannot parse the array's metadata. When zarr-python *can* parse the metadata, the read takes the ordinary path through zarr, and zarr applies every codec the file declares, `pickle` included. The gate is never consulted.

Whether zarr-python can parse the metadata is controlled by the file. Zarr resolves an object array's dtype from the *first* object codec listed in `filters`. A file that lists a benign object codec before `pickle` parses successfully, so the gate is skipped and the pickle codec still runs during decoding.

## Impact for users

Opening an untrusted Zarr v2 file can execute code from that file, with `allow_pickle` left at its safe default.

Two things make this worse than it first sounds.

Execution happens in the **constructor**, not in `read()`. Building `NWBZarrV2IO(path, mode="r")` loads the file's cached namespaces, which reads a spec array, which decodes it. A user who constructs the object and then inspects it before deciding whether to read has already run the payload.

The files are shared by design. NWB Zarr files are downloaded from DANDI and passed between collaborators as a matter of routine, and the v2 reader exists specifically to open files that came from somewhere else and were written by someone else. Opening a downloaded file is the normal case, not an edge case.

The documented promise is currently false. The class docstring states that pickle-encoded datasets require `allow_pickle=True`, and users will reasonably rely on that.

## Reproduction

Against PR #325 at 97e37d0, on a copy of `tests/unit/helpers/nwb_zarrv2_test.nwb.zarr` with `specifications/hdmf-common/1.8.0/namespace/.zarray` rewritten so `filters` is `[{"id": "vlen-utf8"}, {"id": "pickle", "protocol": 5}]` with `dtype: "|O"`, and chunk `0` replaced by a pickled marker object whose `__reduce__` prints and writes a sentinel file.

```
A:  ArrayV2Metadata.from_dict SUCCEEDS for [vlen-utf8, pickle] |O
A2: _v2_array_metadata_supported_by_zarr -> True

    NWBZarrV2IO(store, mode="r")            # allow_pickle left at default False
    PICKLE_EXECUTED
    __init__ returned. sentinel_exists_after_init=True
```

Controls confirm the gate is not simply broken, it has a specific ordering hole:

```
pickle-only / pickle-first:
  zarr_supported=False  (ValueError: No Zarr data type found that matches
                         {'name': '|O', 'object_codec_id': 'pickle'})
  opened via NWBZarrV2IO -> sentinel=False        # pickle did not run

direct gate check:
  _v2_codec(pickle, allow_pickle=False) raised UnsafePickleCodecError
```

So the gate works for the filter ordering the project's own fixture generator happens to produce, and does not work for the ordering an attacker would choose.

## Root cause

Line numbers are from PR #325 at 97e37d0.

`src/hdmf_zarr/backend_zarrv2.py:608` treats "zarr-python can parse this array's metadata" as a proxy for "this array is safe to open":

```python
if not self._v2_array_metadata_supported_by_zarr(zarray_meta):
    ...  # gated fallback decoder, where _v2_codec enforces allow_pickle
```

When the check passes, the read goes to `zarr_obj[entry]` (`src/hdmf_zarr/backend_zarrv2.py:611`), or for spec arrays to `self._group[path][0]` (`src/hdmf_zarr/utils.py:524`), and zarr applies all declared filters through numcodecs, reaching `numcodecs.Pickle.decode` and therefore `pickle.loads`.

The two are not equivalent because zarr resolves an object array's dtype from the first object codec id found in filter order, while decoding applies every filter. `ArrayV2Metadata.from_dict` succeeds on `[vlen-utf8, pickle]` and the pickle filter still runs.

The gate itself (`_v2_codec` / `UnsafePickleCodecError`, `src/hdmf_zarr/backend_zarrv2.py:32-41`) is correct. It is simply not on the path that executes.

`ZarrV2SpecReader._read` falls back to the gated decoder only on `ValueError` or `TypeError` (`src/hdmf_zarr/backend_zarrv2.py:147`), which never fires here because zarr succeeds.

The constructor path is `NWBZarrV2IO.__init__` -> `_build_nwb_manager` (`src/hdmf_zarr/nwb.py:34-44`) -> `ZarrV2IO.load_namespaces` -> `ZarrV2SpecReader._read` -> `super()._read` -> `self._group[path][0]`.

## Suggested fix

Enforce the policy on the file's declared codecs, before handing the array to zarr, rather than inferring safety from whether zarr accepts the metadata.

Scan the entire `filters` list (and the compressor) of every `.zarray` for a pickle codec id and raise `UnsafePickleCodecError` when one is present and `allow_pickle` is `False`, regardless of position and regardless of whether zarr can parse the array. The same check belongs on the spec-reading path, so it applies during `load_namespaces` and therefore during the constructor.

Tests should cover pickle in a non-first filter position, and pickle alongside each of the other object codecs the reader accepts, so this cannot regress to depending on codec ordering again.

rly:
Fixes #385.

## The problem

The `allow_pickle` gate guarded only the raw chunk fallback decoder, which runs when zarr cannot parse an array's metadata. Zarr resolves an object array's dtype from the **first** object codec in `filters` but decodes with **all** of them, so a `.zarray` with `"dtype": "|O"` and `"filters": [{"id": "vlen-utf8"}, {"id": "pickle"}]` takes the ordinary zarr path and reaches `pickle.loads` on the file's own bytes with `allow_pickle=False`.

For a cached spec this happens inside `NWBZarrV2IO.__init__`, via `load_namespaces`, before the caller gets the object back. `_load_namespaces` also turned every exception, including `UnsafePickleCodecError`, into a warning and opened the file anyway.

## The fix

After zarr opens an array, `array.metadata` holds the codecs it will decode with, whether they came from `.zarray` or `.zmetadata`. The policy checks that metadata wherever a v2 array is opened:
- `_iter_children`, for arrays in the groups being read.
- A `ZarrV2IO.resolve_ref` override, for link and reference targets, which can sit where group iteration never goes (for example inside another array's directory).
- `ZarrV2SpecReader`, which refuses pickle whatever `allow_pickle` says. `ZarrSpecWriter` has written cached specs with the JSON codec in every release, so a pickled spec did not come from hdmf-zarr, and legacy ecephys files that need `allow_pickle=True` for `electrodes/group` stay protected during construction.

An array zarr cannot open goes to the raw chunk fallback, which refuses pickle per declared codec and now names the dataset in its error. `_load_namespaces` re-raises `UnsafePickleCodecError`.

`allow_pickle` is removed from `ZarrV2SpecReader`, `ZarrV2IO.load_namespaces`, `_load_namespaces`, and `_build_nwb_manager`, where nothing read it.

## Verification

With a benign payload (`os.makedirs` on a sentinel path) in the crafted chunk, `allow_pickle=False`:

```
                                                  before     this branch
data array, .zarray [vlen-utf8, pickle]           ran=True   UnsafePickleCodecError, ran=False
data array, pickle only in .zmetadata             ran=True   UnsafePickleCodecError, ran=False
cached spec, [vlen-utf8, pickle]                  ran=True   UnsafePickleCodecError, ran=False
link to a pickled array under specifications      ran=True   UnsafePickleCodecError, ran=False
```

The cached spec is also refused with `allow_pickle=True`. The fixture's pickled `electrodes/group` column still reads with `allow_pickle=True`.

Tests cover a pickled group member, a link to an array group iteration never visits, a pickled cached spec under `allow_pickle=True`, and reading the fixture without `allow_pickle`. Removing each check (in `_iter_children`, `resolve_ref`, the spec reader, and the `_load_namespaces` re-raise) fails its test with `UnsafePickleCodecError not raised`, and dropping the dataset name from the fallback error fails the opt in test. The re-raise of `UnsafePickleCodecError` in `_iter_children` has no test: without it, the fallback decoder refuses a `.zarray` that declares pickle on its own, and decodes a `.zmetadata`-only pickle array with the benign `.zarray` codecs.

With `zarr-v3-migration` (including #387 and #390) merged in, `pytest tests/` passes: 606 passed, 1 skipped, with zarr 3.3.0. `ruff` passes. `black` is skipped on these commits because it reformats unrelated blocks in `tests/unit/test_zarrv2_backward_compat.py`, `nwb_zarrv2.py`, and `zarr_utils.py`; the code this PR changes is black clean.


src/hdmf_zarr/backend_zarrv2.py
```diff
@@ -153,6 +165,27 @@ def _read(self, path):
             s = str(s) if not isinstance(s, str) else s
             return json.loads(s)
 
+    def __reject_pickled_spec(self, path):
+        """Refuse a cached spec that declares the pickle codec.
+
+        hdmf-zarr writes cached specs with the JSON object codec, so a pickle codec on
```
oruebel, line 189: To be honest, I am not sure if hdmf-zarr always used JSON code for the spec or whether that was changed at some point. We could relax this to honor the `allow_pickle` argument here as well if this should become an issue. I think for now, it is fine to keep as is for additional safety and adjust in case this should come up as an issue.

oruebel (approved):

src/hdmf_zarr/backend_zarrv2.py
rly: I had AI do a scan. Every hdmf-zarr release from 0.1.0 to 0.13.0 writes specs via `ZarrSpecWriter` with JSON object codec, never pickle.

In any case, I agree. Better to be strict for safety just in case and we can adjust if needed.

rly:
I made some changes to address an edge case to cover links and references and to clean up load_namespaces calls. I don't expect this situation to come up, but `pickle` is a security concern, and it doesn't hurt too much to incorporate these checks.

oruebel (approved):

Issue #277, opened 2025-05-14
[Bug]: Writing scalar dataset with compound dtype


stephprince:
### What happened?

See comment in https://github.com/hdmf-dev/hdmf-zarr/pull/276#pullrequestreview-2841511073 for reference.

This had come up recently in hdmf, but there seems to also be an error with hdmf-zarr writing scalar datasets with a compound dtype.

### Steps to Reproduce

```shell
from pynwb import NWBFile
from uuid import uuid4
from datetime import datetime

from hdmf_zarr import NWBZarrIO

nwbfile = NWBFile(
    session_description="no description.",
    session_start_time=datetime.now(),
    identifier=str(uuid4()),
)

device = nwbfile.create_device(name="Device", description="no description.")

electrode_group = nwbfile.create_electrode_group(
    name="ElectrodeGroup",
    description="no description.",
    device=device,
    location="unknown",
    position=(1.0, 2.0, 3.0),
)

with NWBZarrIO('test.nwb', 'w') as io:
    io.write(nwbfile)  # errors here
    
with NWBZarrIO('test.nwb', 'r') as io:
    read_nwbfile = io.read()
    
    print(read_nwbfile.electrode_groups['ElectrodeGroup'].position.shape)
```

### Traceback

```shell
File "/Users/smprince/.vscode/extensions/ms-python.debugpy-2025.8.0-darwin-arm64/bundled/libs/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py", line 127, in _run_module_code
    _run_code(code, mod_globals, init_globals, mod_name, mod_spec, pkg_name, script_name)
  File "/Users/smprince/.vscode/extensions/ms-python.debugpy-2025.8.0-darwin-arm64/bundled/libs/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py", line 118, in _run_code
    exec(code, run_globals)
  File "/Users/smprince/Documents/code/hdmf-zarr/docs/issue_testing.py", line 24, in <module>
    io.write(nwbfile)
  File "/Users/smprince/Documents/code/hdmf/src/hdmf/utils.py", line 578, in func_call
    return func(args[0], **pargs)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/smprince/Documents/code/hdmf-zarr/src/hdmf_zarr/backend.py", line 347, in write
    super(ZarrIO, self).write(**kwargs)
[...]

PR #391, opened 2026-09-17
Store scalar datasets as zero-dimensional arrays

rly:
Fixes #309. Fixes #277.

## Motivation

`ZarrIO` stored a scalar dataset as a one-element array with shape `(1,)` and marked it with `_SCALAR = true`. `HDF5IO` stores a scalar as a zero-dimensional array with shape `()`. The difference is one more deviation to resolve before `ZarrIO` can merge into core HDMF, and it caused three bugs on this branch:

1. **Scalars read back as one-element arrays.** `__read_dataset` read a scalar with `zarr_obj[()]`, which on a shape `(1,)` array returns the whole array. A dataset written as `5` read back as `array([5])`, and `"hi"` read back as `array(['hi'], dtype=object)`. `ObjectMapper` tolerated this, so the round-trip tests passed.
2. **A scalar object reference read back as a one-element dataset of references.** `HDF5IO` returns a `ReferenceBuilder`.
3. **Writing a compound scalar raised `IndexError`** (#277). Any NWB file with an `ElectrodeGroup` that has a `position` fails to write.

### Why `(1,)` and `_SCALAR` existed

`(1,)` dates to the first `ZarrIO` commit (`94ac871`, April 2022), ported from the hdmf zarr prototype (hdmf-dev/hdmf#98, #696). No commit, PR discussion, review comment, or doc gives a reason. zarr-python has supported zero-dimensional arrays since 2.2.0 (March 2018), before the prototype existed. I confirmed zarr 2.18 writes and reads shape `()` with numeric, JSON and Pickle object codec, and compound data. I could not build the numcodecs versions that zarr 2.4 and 2.11 require on Apple Silicon, so an old object codec bug is not ruled out, but I found no zarr issue reporting one.

`_SCALAR` only exists because of `(1,)`: a one-element array looks exactly like a one-element dataset, so a marker was needed. Shape `()` is unambiguous. #335 §5 adopted LINDI's `(1,)` layout and argued for a boolean marker over `zarr_dtype = "scalar"`, but never considered `()`. `_SCALAR` was introduced in `51785d2` and is in no release, so this PR drops it from reading as well as writing.

## What changed

- **Plain scalars** are written with shape `()` and carry `_DTYPE` (e.g. `"int"`, `"str"`), as arrays already do. Without `_DTYPE`, reading a scalar warns that its dtype was inferred.
- **Scalar object references** are written with shape `()` and read back as a `ReferenceBuilder` pointing at the target builder.
- **Compound scalars** are written by `__scalar_fill__` when the data is zero-dimensional, as in `HDF5IO`. The fixed-length string sizing moves out of `__list_fill__` into `__size_compound_string_fields__`, which handles both a single value and rows. Compound scalars with reference fields are not supported, which matches `HDF5IO`.
- **Compound scalars with a reference field** are not supported, and reading or writing one raises `NotImplementedError` naming the dataset.
- **Compound rank** on the write path is determined with `get_data_shape`, which inspects the first element along each dimension, so a compound dataset supplied as a list of tuples is not materialized to find its rank.
- **Read:** a dataset is a scalar when `zarr_obj.ndim == 0`. `_SCALAR` is no longer a reserved attribute. The `zarr_dtype == "scalar"` branch stays, since `ZarrV2IO` sends v2 arrays that zarr v3 can open through `ZarrIO.__read_dataset`.
- **Cached specs** are written with shape `()`, read back through `ZarrSpecReader._read_spec_string`. A Zarr v2 file stores each spec in a one-element array written with the JSON object codec, which zarr v3 cannot open, so `ZarrV2SpecReader` decodes those from the raw chunks and is unaffected.
- `storage.rst` and the changelog entry for the convention are updated, and there is a Fixes entry for #277.
- The `black` pre-commit hook reformats two existing blocks in `test_zarrv2_backward_compat.py` (a parenthesized `with` and a list comprehension). No behavior change.

## Not in this PR

- **#335 §5** still specifies `(1,)` + `_SCALAR` and needs updating. Noted in [a comment on #335](https://github.com/hdmf-dev/hdmf-zarr/issues/335#issuecomment-5726004776). Zindi adopts shape `()` in [bendichter/zindi#7](https://github.com/bendichter/zindi/pull/7); until its pin here advances, `tests/unit/test_zindi_interop.py` is a strict `xfail`.
- **Scalar object references in `HDF5IO`.** `HDF5IO` cannot read one at all: `h5tools.py` asks a `ReferenceBuilder` for a `dtype` it does not have (hdmf-dev/hdmf#1584). The behavior here follows what that backend intends.
- **Cyclic scalar group references.** A scalar reference to an ancestor group recurses on read, because `__read_group` registers the builder only after descending. No schema produces this, and it needs hand-built builders to reach.
- **Sharding on a scalar.** Scalar writes still pass `io_settings`, as `HDF5IO` does. On zarr 3.3.0 (our minimum), `shards="auto"` on a 0-d array hangs when `array.target_shard_size_bytes` is set (zarr-developers/zarr-python#4304, fixed in 3.4.0). That only happens if a user wraps a scalar in `ZarrDataIO` with sharding.

## How to test the behavior?

```python
import numpy as np, zarr
from hdmf.build import GroupBuilder, DatasetBuilder, ReferenceBuilder
from hdmf_zarr import ZarrIO

target = DatasetBuilder("target", np.arange(3))
root = GroupBuilder("root", datasets={
    "target": target,
    "num": DatasetBuilder("num", 5),
    "txt": DatasetBuilder("txt", "hi"),
    "ref": DatasetBuilder("ref", ReferenceBuilder(target), dtype="object"),
})
with ZarrIO("scalars.zarr", mode="w") as io:
    io.write_builder(root)
g = zarr.open_group("scalars.zarr", mode="r")
print({k: g[k].shape for k in ("num", "txt", "ref")})  # all ()
with ZarrIO("scalars.zarr", mode="r") as io:
    b = io.read_builder()
    print(repr(b["num"].data), repr(b["txt"].data), type(b["ref"].data).__name__)
    # np.int64(5) 'hi' ReferenceBuilder
```

The #277 reproduction (an `ElectrodeGroup` with `position=(1.0, 2.0, 3.0)` written with `NWBZarrIO`) now writes, and `position` reads back as `array((1., 2., 3.), dtype=[('x', '<f4'), ('y', '<f4'), ('z', '<f4')])`, the same value `NWBHDF5IO` returns.

New tests: `TestScalarDataset` (plain, reference, compound, and the two unsupported compound-with-reference cases) in `test_zarrio.py`. Each was checked by mutation, and each mutation made a test fail:

| Mutation | Failure |
|---|---|
| Read returns the 0-d array without unwrapping | `test_numeric_and_string`: data is a `zarr.Array` |
| Scalar reference not resolved | `test_reference`: `'/target'` is not a `ReferenceBuilder` |
| Scalar reference written as `(1,)` | `test_reference`: shape `(1,) != ()` |
| Compound scalar sent to `__list_fill__` | `test_compound`: the #277 `IndexError` |
| Compound scalar written without string sizing | `test_compound`: zarr rejects a zero-length string field |
| Sizing ignores the data (always 512) | `test_compound`: 600-character value truncated to 512 |
| Write guard removed | `test_compound_with_reference_field_write`: `TypeError: iteration over a 0-d array` escapes |
| Read guard removed | `test_compound_with_reference_field_read`: the reference reads back as the string `'/target'` |

Full suite: 610 passed, 1 xfailed. `ruff` and `codespell` pass.

## Checklist


oruebel:
> `(1,)` dates to the first `ZarrIO` commit (`94ac871`, April 2022), ported from the hdmf zarr prototype ([hdmf-dev/hdmf#98](https://github.com/hdmf-dev/hdmf/pull/98), #696). No commit, PR discussion, review comment, or doc gives a reason.

I am not sure, but I think this was at the time consistent with how h5py / HDF5IO handled scalars at the time. I think representation of scalars as numpy scalar arrays rather than as Python primitive scalars was introduced later. 

> `_SCALAR` only exists because of `(1,)`: a one-element array looks exactly like a one-element dataset, so a marker was needed. Shape `()` is unambiguous.

I agree, eliminating custom handling of scalar types will be useful.

rly:
@oruebel @bendichter this is ready for review and I believe this is the last PR for #325.

bendichter:
Conceptually it looks good to me

oruebel (approved):

PR #393, opened 2026-09-20
Fix Zarr v2 write guard, allow_pickle discoverability, and the renamed compressor argument

rly:
Four fixes found while reviewing #325, plus a changelog entry. One commit each.

**Refuse a Zarr v2 hierarchy opened for writing** — the v2 hint was gated on read modes, so opening a v2 file for append gave a raw zarr parse error. Worse, a v2 group whose metadata zarr can parse opened without error, and the write then emitted Zarr v2 output into the legacy hierarchy, contradicting `README.rst` and `overview.rst`. Now refused in every mode, keyed on the format of the hierarchy actually opened.

**Name `allow_pickle` where the Zarr v2 conversion path is documented** — the flag appeared nowhere in `docs/` or `README.rst`, yet most Zarr v2 NWB files carry a pickle-encoded dataset, so the `convert_to_v3` call the v3 backend prints aborts when followed verbatim. Message and docs now name it. Default stays `False`.

**Reject the renamed `compressor` argument with a pointed error** — it got docval's `unrecognized argument`, which reads as a typo of a valid argument and says nothing about the accepted type having changed too. Now rejected with a message naming both.

**Bump the zindi pin to the zero-dimensional scalar convention** — zindi stores HDF5 scalars as zero-dimensional zarr arrays now, so the strict xfail reports XPASS and the mark comes off. Verified in a clean venv: old pin fails with `JSONDecodeError`, new pin passes.

615 passed, 1 skipped locally (zindi skips when it is not installed).


src/hdmf_zarr/backend.py
```diff
@@ -301,6 +301,7 @@ def open(self):
                 # error. Point the user at the Zarr v2 backend instead.
                 self._raise_if_zarr_v2(e)
                 raise
+            self._raise_if_opened_zarr_v2()
```
oruebel: Can you clarify. The comment says that `Opening a Zarr v2 file with the Zarr v3 backend fails here` so if that is true then we would have already raised an error. However, then this calls `self._raise_if_opened_zarr_v2()` after we successfully opened the file to check again? In what case would we successfully open a V2 file but then need to check after opening?

If we need to check after opening the file, the shouldn't `self._raise_if_opened_zarr_v2()` always be called rather than only `if self.__file is None:`?

`_raise_if_opened_zarr_v2()` is only called here, i.e., I am wondering whether it may be clearer to move the code here directly.

oruebel: I think this is probably mainly a documentation issue rather than an error

oruebel (approved):

src/hdmf_zarr/backend.py
rly: I had wondered the same thing when the AI identified this hole. I should have left a comment / amended the above comment to make it clear.

`_open_file` or `_open_file_consolidated` raises an `Exception` that is caught above this, when opening a v2 file that has metadata that zarr-python v3 can't map, such as object-dtype arrays with `pickle`/`json2`/`vlen-utf8` codecs or an int `fill_value`. A non-consolidated-metadata v2 file gets past the `try/except` and has `zarr_format=2`. It then raises an error on `read()`. We can raise that error earlier during `__init__`-> `open()` instead. See examples at the end of this comment. I'll update the inline comments to make it more clear.

> shouldn't `self._raise_if_opened_zarr_v2()` always be called rather than only `if self.__file is None`

`self.__file` is assigned to a non-None value only in this `open()` function. So, right now `_raise_if_opened_zarr_v2` is called at the only time `self.__file` is set. We could move it to outside the `if` statement, but then a second `open()` call would call an extraneous `_raise_if_opened_zarr_v2`. Doesn't really hurt, but it's not necessary.

>`_raise_if_opened_zarr_v2()` is only called here, i.e., I am wondering whether it may be clearer to move the code here directly.

I agree. I'll make that change.

Example to generate a zarr v2 file that passes the v3 branch's `NWBZarrIO.open()`:
```
uv venv --python 3.11 /tmp/v2env
VIRTUAL_ENV=/tmp/v2env uv pip install "hdmf-zarr<0.14" pynwb
```
```python
from datetime import datetime
from pynwb import NWBFile

nwb = NWBFile(session_description="legacy v2, not consolidated", identifier="legacy-1",
              session_start_time=datetime(2024, 1, 1, tzinfo=tzlocal()))
with NWBZarrIO("/tmp/legacy_v2.nwb.zarr", mode="w") as io:
    io.write(nwb, consolidate_metadata=False)
```

Here's the result of different calls. I could see someone creating a shared `manager` and passing it to two `NWBZarrIO` for read and export, or to a single `NWBZarrIO` for append, but admittedly these are unlikely scenarios.

| what you run | without this PR | with this PR |
| --- | --- | --- |
| `NWBZarrIO(p, "r")` or `NWBZarrIO(p, "a")`, then `.read()` | `ValueError` with helpful hint, raised by `load_namespaces` during `__init__()` | unchanged |
| `NWBZarrIO(p, "r", manager=...)`, then `.read()` | constructor returns a `zarr_format=2` handle, then `read_builder` raises a `ValueError` with helpful hint | `ValueError` with helpful hint raised sooner, at `__init__()` -> `open()` |
| `NWBZarrIO(p, "a", manager=...)`, then `.read()` | constructor returns a `zarr_format=2` handle, then the read raises `TypeError: Invalid type: 0. Expected a string.` | `ValueError` with helpful hint raised sooner, at `__init__()` -> `open()` |

I thought there were other worse edge cases, but it turns out they were either unrealistic or didn't get past `read()`.

Issue #394, opened 2026-09-20
Text datasets are written one element at a time, making writes superlinear in length


rly:
Found while reviewing #325. Applies to the `zarr-v3-migration` branch, not a released version.

## What happens

`ZarrIO.__list_fill__`'s guard changed from `if dtype == object:` to [`if dtype == str:`](https://github.com/hdmf-dev/hdmf-zarr/blob/1a905e565e3350030336a6122b8499aea3dcae9e/src/hdmf_zarr/backend.py#L1802). Because `__dtypes["text"]` is the Python `str` type, every ordinary variable-length text dataset now takes the per-element `for c in np.ndindex(data_shape): dset[c] = o` loop instead of one vectorized `dset[:] = np.array(data, dtype=dtype)`.

`VariableLengthUTF8` does not implement `HasItemSize`, so zarr's chunk guess uses `item_size=1` and clamps to a 128 KiB target, which puts any text array up to roughly 196k elements in a **single chunk**. The array-to-bytes codec for vlen UTF-8 is not an `ArrayBytesCodecPartialEncodeMixin`, so each `dset[c] = o` falls into `_async_write_fallback` and fetches, decodes, merges, re-encodes and re-stores the entire chunk. N assignments against one N-element chunk.

## Measurements

Writing one text `DatasetBuilder` on `1a905e5`:

| n | time | chunks |
|---|---|---|
| 1,000 | 1.24 s | (1000,) |
| 2,000 | 2.73 s | (2000,) |
| 4,000 | 7.41 s | (4000,) |
| 8,000 | 23.1 s | (8000,) |
| 16,000 | 64.6 s | (16000,) |

A single vectorized assignment into an equivalent array at n=16,000 is **0.0027 s**, roughly 24,000x faster. Extrapolating, a 1e6-element text column (a `VectorData` of text, `Units.quality`, `ImageSeries.external_file`) is hours, and presents to the user as a hung write with no progress output. Data that does land is correct; the failure mode is an apparent hang, and a truncated file if the user kills it.

## Impact

Hits every new user writing an NWB file with a non-trivial text column. No existing test covers it, because the suite only writes small datasets.

## Cause confirmed, and a validated fix

Reverting the guard to `if dtype == object:` in a scratch worktree drops n=8,000 from 24.4 s to 0.0055 s and n=16,000 from 63.4 s to 0.0097 s, **and the full suite stays green** (611 passed / 1 skipped, identical to baseline).

[...]

PR #397, opened 2026-09-20
Write a text dataset in one assignment

rly:
Fixes #394.

`__list_fill__` resolves a text dataset to the Python `str` type and wrote it one element at a time. Zarr sizes a variable-length string array to a single chunk up to roughly 196k elements, and its UTF-8 codec has no partial encoder, so each assignment decoded, merged, re-encoded and re-stored the whole chunk.

| n | before | after |
|---|---|---|
| 1,000 | 1.24 s | 0.007 s |
| 4,000 | 7.41 s | 0.005 s |
| 16,000 | 64.6 s | 0.007 s |
| 64,000 | — | 0.014 s |

`StringDType` decodes bytes as UTF-8, so a single `np.array` call covers the `str` and `bytes` cases the loop handled. I checked the two paths agree element for element on str, bytes, mixed, 2-D str, 2-D bytes, ints, empty strings and non-ASCII unicode before relying on that.

Bytes that are not valid UTF-8 are the one thing the bulk write cannot express. `UnicodeDecodeError.object` carries the value that failed to decode, so the message from #396 that names the dataset and the value is built straight from the exception rather than by visiting each element to find the culprit. `_invalid_utf8_error` holds that message for both this path and `_decode_for_text_dataset`, and `TestNonUTF8BytesDataset` covers it.

Every other conversion failure reaches the caller. The case behind that choice is a ragged nested list: `get_data_shape` reports a rectangular shape for one, `numpy>=2` refuses to build the array, and `HDF5IO` raises. 0.13.0 raised as well, because a text dataset took the vectorized branch there, so this is the behavior of the released version rather than a new one.

619 passed, 1 skipped.

**Scope note**: #394 also mentions the same pattern for object reference datasets, which write reference paths one index at a time in `__write_dataset__`. That site is left alone here — smaller N in practice, and it writes reference paths rather than user text, so it deserves its own change. (The line numbers in #394 are pinned to `1a905e5` and no longer match this branch.)


src/hdmf_zarr/backend.py
```diff
@@ -1823,11 +1823,16 @@ def __list_fill__(self, parent, name, data, options=None):  # noqa: C901
...
-            for c in np.ndindex(data_shape):
-                o = data
-                for i in c:
-                    o = o[i]
-                dset[c] = _decode_for_text_dataset(o, name, parent.name)
+            try:
+                # StringDType decodes bytes as UTF-8, so one array covers str and bytes alike
+                dset[...] = np.array(data, dtype=zarr_dtype)
+            except (ValueError, TypeError):
+                # A source that numpy cannot convert in a single call is written element by element
```
oruebel: Is this an unexpected error case? If so should we raise a warning here?

tests/unit/test_zarrio.py
```diff
@@ -850,3 +850,38 @@ def test_utf8_bytes_are_written_as_text(self):
...
+        """
+        n = 500
+        values = ["label_%04d" % i for i in range(n)]
+        builder = GroupBuilder("root", datasets={"txt": DatasetBuilder("txt", values, dtype="text")})
+
+        original = zarr.Array.__setitem__
+        calls = []
+
+        def counting_setitem(self, key, value):
+            calls.append(key)
```
oruebel: If we raised a warning when not writing in one pass, then that would also simplify this test by allowing to check for the warning rather than having to count calls and also allow detection of this behavior in other tests rather than passing silently

src/hdmf_zarr/backend.py
rly: No. Sorry, I was still digging into this when you wrote this. 

Ragged lists and non-UTF-8 numpy arrays with S dtype were the culprits, but since we now require `numpy>=2`, `np.array` catches ragged lists, and non-UTF-8 data is highly unlikely and we don't need to iterate through the array to find the bad one. Let's keep it simple.

tests/unit/test_zarrio.py
rly: Removed. `numpy.array` will raise an error for ragged lists and `UnicodeDecodeError` for non-utf8 byte arrays. No more per-element assignment. No need for a test IMO.

oruebel (approved):

PR #398, opened 2026-09-21
Translate the on-disk reference dtype name once, at read time

rly:
## Summary

Our Zarr v3 convention stores `"object_reference"` in `_DTYPE` for reference datasets. The read path translated it to hdmf's `DatasetBuilder.OBJECT_REF_TYPE` (`"object"`) for plain reference datasets but not for compound fields listed in `_REFERENCE_FIELDS`, so `_is_ref`, the compound read checks, and `AbstractZarrTableDataset` all accepted both names via a string literal "object_reference" repeated several times across two files.

- Translate compound reference fields to `OBJECT_REF_TYPE` at read time as well, and compare against `OBJECT_REF_TYPE` everywhere downstream.
- Keep the on-disk name in one documented constant, `REFERENCE_DTYPE_ATTR_VALUE`.
- Derive `_REFERENCE_FIELDS` on write from the reference indices already computed, dropping the intermediate `type_str` list.
- Remove the unused `"object_reference"` entry from `ZarrIO.__dtypes`.

The file format is unchanged. Existing tests cover both read-side translations (reverting either one fails `test_read_reference_compound`, `test_read_reference`, and the Zarr/HDF5 export roundtrips).


src/hdmf_zarr/backend.py
```diff
@@ -979,7 +988,7 @@ def _is_ref(self, dtype):
         elif isinstance(dtype, np.dtype):
             return False
         else:
-            return dtype in (DatasetBuilder.OBJECT_REF_TYPE, "object_reference")
+            return dtype == DatasetBuilder.OBJECT_REF_TYPE
```
oruebel: In this particular `_is_ref` function is think it would be Ok to leave the check to also check for `REFERENCE_DTYPE_ATTR_VALUE`

```suggestion
            return dtype in (DatasetBuilder.OBJECT_REF_TYPE, REFERENCE_DTYPE_ATTR_VALUE)
```

src/hdmf_zarr/zarr_utils.py
```diff
@@ -155,7 +155,7 @@ def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.__refgetters = dict()
         for i, t in enumerate(types):
-            if t in (DatasetBuilder.OBJECT_REF_TYPE, "object_reference"):
+            if t == DatasetBuilder.OBJECT_REF_TYPE:
```
oruebel: Should this and below also use the `_is_ref(` function to check for reference type, mainly to centralize this kind of check?

oruebel (approved):
A couple of minor comments, but nothing critical.

src/hdmf_zarr/backend.py
oruebel: If we don't want to check for `REFERENCE_DTYPE_ATTR_VALUE` here then it would be useful to clarify this in the docstring to avoid questions later on

src/hdmf_zarr/zarr_utils.py
rly: Good idea

rly: Refactored in fd0704d

src/hdmf_zarr/backend.py
rly: `REFERENCE_DTYPE_ATTR_VALUE` shouldn't ever reach this function. In the refactor, I added a note in the docstring.

src/hdmf_zarr/zarr_utils.py
oruebel: Nice

PR #400, opened 2026-09-21
Remove code for zarr<3.3 and fix issues found along the way

rly:
## Motivation

The minimum zarr version is now 3.3.0. This removes code that existed only for zarr 3.0 to 3.2 or zarr 2, and fixes problems found while auditing it.

## Changes

- Remove warning filters that match nothing under zarr 3.3.0: the numcodecs spec warning in `hdf5_to_zarr_filters` and the gallery runner (emitted only by zarr 3.1.x), the unstable-dtype gallery filter (zarr 3.3.0 no longer emits it for fixed-length unicode or structured dtypes), and the `NestedDirectoryStore` and numpy `__array__` copy gallery filters (zarr 2).
- Fix the S3 streaming gallery example never reading its file under `test_gallery.py`. The gallery runner turns warnings into errors. `ZarrV2IO` issued its fallback-read warning inside the `try` that guards the fallback decode, so the raised warning was treated as a failed decode and the entry was recorded as skipped. The skip warning then aborted the read, and the example's bare `except Exception` printed a note while the gallery reported success. The fallback-read warning is now issued after a successful decode, the gallery runner ignores that expected warning, and the example catches only `ImportError` (fsspec missing) and `OSError` (network unavailable).
- Make the missing-fsspec error reachable. `from zarr.storage import FsspecStore` succeeds without fsspec, so `FSSPECSTORE_AVAILABLE` was always `True` and users got a bare `ModuleNotFoundError: No module named 'fsspec.core'`. `_resolve_store` now checks with `importlib.util.find_spec("fsspec")` for `storage_options` or a protocol URL and raises an `ImportError` naming `hdmf-zarr[full]`. `SUPPORTED_ZARR_STORES` always includes `FsspecStore`.
- Remove `DISABLE_ZARR_COMPRESSION_TESTS`, which was always `False` because `BloscCodec` and `TransposeCodec` exist in every zarr 3 release.
- Update docs that described zarr v2: `.zattrs` and `.zmetadata` are now `zarr.json`, `zarr.convenience.*` is now `zarr.open` / `zarr.consolidate_metadata`, and the consolidated metadata link points at the current zarr docs.

## Testing

- `test_fallback_read_warning_as_error_propagates` and `TestRemoteStoreWithoutFsspec` each fail on their assertion with the corresponding fix reverted.
- Full test suite passes with zarr 3.3.0. The gallery passes and the S3 example now prints the file's session description. With fsspec blocked, the S3 example prints the install hint and continues.


docs/gallery/plot_s3_streaming.py
```diff
@@ -62,8 +62,9 @@
         print(f"Session Description: {nwbfile.session_description}")
         print(f"Identifier: {nwbfile.identifier}")
         print(f"Subject ID: {nwbfile.subject.subject_id if nwbfile.subject else 'N/A'}")
-except Exception as e:
-    print(f"Note: Could not access S3 file (network access may be required): {e}")
+except (ImportError, OSError) as e:
+    # fsspec is not installed or the network is unavailable
+    print(f"Note: Could not access S3 file: {e}")
```
oruebel:
```suggestion
    # fsspec is not installed or the network is unavailable
    # `pip install hdmf-zarr[full]` installs zarr with fsspec
    print(f"Note: Could not access S3 file: {e}")
```

docs/gallery/plot_s3_streaming.py
```diff
@@ -170,8 +172,9 @@
 try:
     nwbfile = io_class.read_nwb(s3_url)
     print(f"Session Start Time: {nwbfile.session_start_time}")
-except Exception as e:
-    print(f"Note: Could not access S3 file (network access may be required): {e}")
+except (ImportError, OSError) as e:
+    # fsspec is not installed or the network is unavailable
+    print(f"Note: Could not access S3 file: {e}")
```
oruebel:
```suggestion
    # fsspec is not installed or the network is unavailable
    # `pip install hdmf-zarr[full]` installs zarr with fsspec
    print(f"Note: Could not access S3 file: {e}")
```

oruebel (approved):

PR #403, opened 2026-09-22
Raise docs matplotlib floor to 3.8.4

rly:
## Motivation

The `macos-gallery-python3.12-minimum` job in "Run all tests" fails on `dev` ([run](https://github.com/hdmf-dev/hdmf-zarr/actions/runs/35676217089)). With `--resolution lowest-direct`, the docs group's `matplotlib>=3.6` resolves to 3.6.0. That version has no Python 3.12 wheel, so it is built from source, and the bundled freetype 2.6.1 fails to compile on the macOS runner (`unknown type name 'Byte'`). The Linux and Windows builds succeed.

matplotlib 3.8.0 through 3.8.3 have 3.12 wheels but declare `numpy<2`, which conflicts with `numpy>=2.0.0`. 3.8.4 is the first release that installs, so this sets the docs floor to `matplotlib>=3.8.4`. It also fixes a typo in README.rst.

This change is needed before the 0.14.0 release. It only touches the `docs` dependency group, which is not part of the published package metadata.

## How to test the behavior?

`tox -e gallery-py312-minimum -r` passes locally with Python 3.12 and installs matplotlib 3.8.4.

## Checklist

- [x] Did you update `CHANGELOG.md` with your changes? Not needed, this only changes a dev dependency group.
- [ ] Does the PR use "Fix #XXX" notation to tell GitHub to close the relevant issue numbered XXX when the PR is merged? No related issue.


rly:
Going to bundle this into a release PR with some other minor fixes.

